PR #15, opened 2019-12-19
Other sensors dev

Magalex2x14:
I think, it's time to release.


Magalex2x14:
**Changes since 0.3.0:**

In addition to LYWSDCGQ, added support for the following Xiaomi sensors (please check updated README.md):
- LYWSD02 (rectangular body, E-Ink)
- CGG1 (round body, E-Ink)
- HHCCJCY01 (MiFlora)

**Attributes added:**

- RSSI in dBm (Received Signal Strength Indicator)
- Device type

**Fixed:**

- Now the component correctly terminates the subprocesses, there are no more zombies when restarting HA.
- Improved malformed (or unknown?) messages detection.

Ernst79:
Great. Will check it in the weekend.

Ernst79:
@Magalex2x14 I've checked the code, blackified the code and fixed some flake8 errors. 

I have one pylint error left on line 315. You have a 'continue' on that line. Pylint is saying:

> sensor.py:315:16: R1724: Unnecessary "else" after "continue" (no-else-continue)

I'm not 100% sure, but I tried by leaving out the 'continue' on line 315, and it runs fine on my end. Can you confirm that we can leave out this line. I haven't changed it here on GitHub, but if you agree, please remove line 315. 

Other  than that, it's fine to make the release.

Magalex2x14:
Yes, "else:" does not make sense here, because if the condition is met, the current iteration of the loop will not continue. Thank you.

PR #33, opened 2020-02-03
No external tools

Magalex2x14:


Ernst79:
I will check it tonight (European time)

Magalex2x14:
Looks good, but I have a question. I always thought that parenthesis when unpacking tuples in python are optional (lines 140-159). You put them based on some kind of requirement (lint or something)?

Ernst79:
The change was made by Black. I normally copy and paste my code on [black.now.sh](black.now.sh). This site applies black formatting which is required for official components in Home Assistant. You can just copy and paste your code and copy the altered code back to github. See also [black website](https://black.readthedocs.io/en/stable/). 

Pylint and flake8 didn't complain about your original code, so both ways are probably correct, as you say. The only problem when making the custom component an official HA integration in the future is that it has to be black compliant, (and also pylint compliant and flake8 compliant). So I thought it is better to do this right now, to safe you some work later.

Magalex2x14:
OK! Thanks for this info!

PR #39, opened 2020-02-21
Cleanup unnecessary intermediate conversions

a1ien:
Use bytes for data manipulation. It's faster and not need convert to string and back to bytes for struct.unpack

### DIFF
--- a/custom_components/mitemp_bt/const.py
+++ b/custom_components/mitemp_bt/const.py
@@ -31,12 +31,13 @@ CONF_HMAX = 99.9
 
 # Xiaomi sensor types dictionary with offset for adv parser
 XIAOMI_TYPE_DICT = {
-    '209800': ["HHCCJCY01", 1],
-    '20AA01': ["LYWSDCGQ", 0],
-    '205B04': ["LYWSD02", 1],
-    '304703': ["CGG1", 0],
-    '205D01': ["HHCCPOT002", 1]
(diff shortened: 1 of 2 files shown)


Magalex2x14:
Hurrah! Finally, a man appeared who drew attention to this :)
Yes, I know, but there were many other tasks, and my hands did not reach. Thank you!

I’ll take a look later.

Magalex2x14:
This continues from the very beginning, from the time when we received strings at the input.

Ernst79:
Look good to me. I’m very limited in time this week, so I don’t have time to do a real live test unfortunately. But I suppose you have checked it on a running HA, @Magalex2x14 . I suggest that you merge it. Thank for your help @a1ien , much appreciated!

Magalex2x14:
There are no errors during execution, but the data does not come. Need debugging.
@a1ien does this code work for you?

a1ien:
Yes. But also I test it with additional patch for NUC.

Magalex2x14:
@a1ien I have discovered a problem. The code works, but for some reason HA created new entities with this code... It remains to understand why.

image

After restoring the old code, the correct entities are updated.

Magalex2x14:
Perhaps some kind of confusion with the types?

Magalex2x14:
Guessed a peek into the .storage/core.entity_registry :)
image

PR #46, opened 2020-03-02
fix typo and device class

Ernst79:
Two small changes. 

Typo fix: xiomi => xiaomi
Comment for def device_class(self): was copied from unit of measurement --> changed to device class


Ernst79:
@Magalex2x14 

Can you have a look? Two (very) minor changes.

PR #74, opened 2020-05-03
New "whitelist" option and Xiaomi Smart Repellent support

Magalex2x14:


Magalex2x14:
Tested. Me and at least two users requesting a whitelist and repellent support.

Magalex2x14:
Thank you!

PR #78, opened 2020-05-22
Remove HASSio from documentation

Ernst79:
Small change in docs, as the name HASSio no longer used and is changed to home assistant (and home assistant to home assistant core).


Magalex2x14:
Oh yes, thanks.

PR #106, opened 2020-08-24
Add new sensor: MHO-C401

znanev:
MHO-C401

There is a new version of a nice square E-Ink temperature and humidity sensor, which now supports Bluetooth. This new sensor is listed as **_MHO-C401 (2020 version)_** [here](https://www.aliexpress.com/item/4001174769598.html).

Its Bluetooth operations are quite similar to those of LYWSD03MMC - i.e. it broadcasts temperature and humidity once in about 10 minutes and battery level once in an hour, and advertisements are encrypted.

I've added this new device (its XIAOMI_TYPE bytes are 0x8703) to the dictionaries in const.py. Then I just had to extract the `bind_key` from the modified Xiaomi Home App in order to supply it to the `encryptors` configuration, and the new sensor appeared in my Home Assistance instance.

Probably the only thing that needs to be modified (code-wise) is the sensors.jpg image, in order to include a picture of the new sensor.

Hope you'll find this Pull Request useful.

Cheers,
zn

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ This custom component is an alternative for the standard build in [mitemp_bt](ht
 - WX08ZM
 
   (Xiaomi Mija Mosquito Repellent, Smart version, broadcasts switch state, tablet resource, battery level, about 50 messages per minute)
+  
+- MHO-C401
+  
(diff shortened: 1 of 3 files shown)


Ernst79 (approved):
Thanks for your contribution. no comments. I'll update the picture and will make a new version later today.

PR #126, opened 2020-10-20
Measuring sensor

Ernst79:
This time, only some code optimization, reducing the number of lines of python code with about 20%.

- Adding a MeasuringSensor class which is used for all measuring sensors, to shorten the python code (won't affect users)
- Adding device_class Illuminance to illuminance sensors (will change the icon to default Home Asssistant icon)
- Using default unit of measurements from Home Assistant constants (won't affect users)

Changes were based on a small part of the changes as proposed in beta 0.7.1 by @Magalex2x14, further developed by me (@Ernst79).

### DIFF
--- a/custom_components/mitemp_bt/sensor.py
+++ b/custom_components/mitemp_bt/sensor.py
@@ -17,7 +17,6 @@ from homeassistant.const import (
     DEVICE_CLASS_ILLUMINANCE,
     DEVICE_CLASS_TEMPERATURE,
     CONDUCTIVITY,
-    LIGHT_LUX,
     PERCENTAGE,
     TEMP_CELSIUS,
     ATTR_BATTERY_LEVEL,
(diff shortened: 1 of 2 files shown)


Ernst79:
@Magalex2x14 now you are back, perhaps you have time to do a review?

Magalex2x14:
Yes, I'll see tomorrow )

Ernst79:
I've reverted LIGHT_LUX import, as this only works on HA core >= 0.116. Let's wait a little while to make sure most users are updated.

Magalex2x14:
We can get HA version, and make import if it is >= 116 :)
What do you think?

Magalex2x14:
I tested your changes with Python 3.7.3. Everything looks good. I will update to 3.8.6 this weekend.

Ernst79:
I don’t think it is worth the effort. It is only another way to import lx as unit of measurement. I will make a separate PR for this which we can merge later, e.g. when HA 0.118 is released. By then, most people will have updated, I assume.

Magalex2x14 (approved):

PR #140, opened 2020-11-02
code cleaning

Ernst79:
Some cleaning of part of the code

### DIFF
--- a/custom_components/mitemp_bt/sensor.py
+++ b/custom_components/mitemp_bt/sensor.py
@@ -610,73 +610,74 @@ def setup_platform(hass, config, add_entities, discovery_info=None):
             if data and "mac" in data:
                 # ignore duplicated message
                 packet = data["packet"]
-                prev_packet = lpacket(mac=data["mac"])
+                mac = data["mac"]
+                prev_packet = lpacket(mac)
                 if prev_packet == packet:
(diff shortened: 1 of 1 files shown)


Ernst79:
@Magalex2x14 This was one of your own code optimizations you had in 0.7.1 beta. I just did it again on the latest code (all credits to you :-).

Magalex2x14:
OK )

Magalex2x14 (approved):

Magalex2x14:
There are many good changes in that branch, but there were also controversial ones. Since I broke that stream with my long absence, now I can't remember all the nuances. I remember one thing - there was a change that spawned the data collection thread only once at system startup. Now we have this thread re-created each time at the beginning of a new measurement period. I think it's better to leave it that way until we learn to catch all possible problems on the hci interface. Otherwise, users who use our component in conjunction with other bt components risk being left without data at all...

Magalex2x14:
If you like, then keep pushing those changes to the current branch. I didn't plan to return to that beta anytime soon. I want to finish reading the missed issues first.

Ernst79:
Yes, sure. I started with the easy changes :-). I propose to make small steps (commits). I also have an idea about the configuration parameters, but will make a separate issue for that to discuss.

PR #147, opened 2020-11-09
[WIP] Component reengineering

Magalex2x14:
The main goal is the possibility of an individual (per device) settings for calculating states and updating the switching sensors in real-time.

ToDo:

- [ ] reworking entity classes, transferring the state calculation procedure to the entity update method
- [ ] split the message parser into a filter and a parser, move the logic for controlling message duplicates to the earliest possible stage
- [ ] perform parser and accumulation of measurements in real-time as new messages arrive


Magalex2x14:
OMG... I tried to make a merge from realtime_tests (just to look into it) and rolled it back...
But the commits still somehow ended up here ...

Ernst79:
That’s a pity. Hopefully there is a way to revert. Maybe it’s easier to start over.

PR #148, opened 2020-11-09
Component reengineering

Magalex2x14:
The main goal is the possibility of an individual (per device) settings for calculating states and updating the switching sensors in real-time.

ToDo:
- [x] solve #61 by fixing and bundling aioblescan with our component
- [x] mccgq02hl #115 
- [x] reworking entity classes, transferring the state calculation procedure to the entity update method
- [x] ~~split the message parser into a filter and a parser, move the logic for controlling message duplicates to the earliest possible stage~~
^postponed, because this way I lose the ability to see the number of HCI events actually processed
- [x] perform parser and accumulation of measurements in real-time as new messages arrive


Magalex2x14:
Yes, it turned out to be easier to start over )

Magalex2x14:
@Ernst79 just to keep you informed - I made the decision to force #61 (there are too many problems with this, and the further it gets, the worse it gets). While I'm doing this, 0.8.1 can be released, after which we will roll out a new beta with a fix for "NUCs" and initial support for mccgq02hl. While the owners of mccgq02hl will try the beta, I will deal with the issue of updating the binary sensors in real-time.

Alternatively, we can continue to publish intermediate beta versions step by step, and make a full-fledged release after all the described changes.

Magalex2x14:
@Ernst79 while I was expanding the sensor classes for MCCGQ02HL, I noticed a nuance that could potentially be a problem - we have non-unique unique_id's:

```python
self._sensor_name = sensor_name(config, mac, "conductivity")
self._unique_id = "c_" + self._sensor_name
```

And here is an entry from the HA entity registry:
```
{
                "entity_id": "sensor.ble_conductivity_flower",
                "config_entry_id": null,
                "device_id": null,
                "unique_id": "c_flower",
                "platform": "ble_monitor",
                "name": null,
                "icon": null,
                "disabled_by": null,
                "capabilities": {},
                "supported_features": 0,
                "device_class": null,
                "unit_of_measurement": "\u00b5S/cm",
                "original_name": "ble conductivity flower",
                "original_icon": "mdi:flash-circle"
},
```

I remember that earlier there was a mac address and an entity type.
As far as I understand, the mac address disappeared from the ID when the entity naming system was changed.
Did you remove the mac address for some reason?

Ernst79:
No, it had not a specific reason to remove it. I think we should move it back to the mac address, you're right. I'll change that in the integration level branch and propose to do both one more test and release it today.

Magalex2x14:
Wait. On the other hand, this approach gives some control over the uniqueness of the sensor name. Have you tried giving the same sensor names? Do you know, how does HA react to non-unique IDs?

Ernst79:
I'll try, give me a few minutes.

Ernst79:
You will get an error in the log

`Platform ble_monitor does not generate unique IDs. ID t_garden already exists - ignoring sensor.ble_temperature_garden
Home Assistant has started!`

It is only updating one of the sensors. Makes sense. I propope to change it back to `"c_" + mac`

Magalex2x14:
I am simply confused by the fact that if we change the ID again, then those who have already updated for the beta will have to deal with the new entities again... If there is no urgent need, then maybe we can leave it as it is? :)

Ernst79:
That's true. Users will be warned anyways. 

I now remember that I have done this to force home assistant to recreate a new sensor, as it would otherwise keep the old name with the mac at the end. The sensor name with mac at the end coupled to the unique ID when initially created without the  `name` option. When you add a `name` option, it will stick to the old name, as it is stored somewhere in the HA entity registry.

Ernst79:
Shall we release 0.8.1 as final now?

Magalex2x14:
I think yes, we can do it.

Ernst79:
Ok, I'll make the merge, remove the word beta and make a release. Will you make an announcement on the forum?

Ernst79:
Done

I'll check the beta 0.8.2 release tonight.

Magalex2x14:
OK, thank you. In fact, I think it is better to test this branch. It works for me.
I will interrupt for now, I will continue either tonight or tomorrow ...

Ernst79:
Ok, will do

Ernst79:
I saw in the door sensor issue that you were looking for a possibility to restore the state on startup. This should be possible with the RestoreEntity option, see this [forum topic](https://community.home-assistant.io/t/main-purpose-of-restoreentity/135270/2). See also this [explanation](https://aarongodfrey.dev/programming/restoring-an-entity-in-home-assistant/). 

You will need to add `from homeassistant.helpers.restore_state import RestoreEntity` at the top of sensor.py and I think you will have to change `MeasuringSensor(Entity)` to `MeasuringSensor(RestoreEntity)` as well, I think.  With the `async_get_last_state()` method you should be able to restore the state.

Magalex2x14:
Hmmm... Thank you. I think this is worth implementing, but not now. Our conversations on that issue are not very relevant, they confused me. The sensor sends data at least once a minute (judging by the dumps), as it should according to the protocol. I'll finish realtime, and then we'll see what he really had there.

Magalex2x14:
Remained a little, looks like )

Ernst79:
Sure, this can be implemented later. I can also have a look at it later, but for now, I let you do your work 👍

Magalex2x14:
Wheoou... 😅
It seems to be working fine. We need to test. Please try this branch, when you have time...

Magalex2x14:
In the morning, when I get up, I'll see what's on the forum.

Ernst79:
Component is running fine. Strange thing is that the review option is now showing differences compared to 0.7.x, so it is impossible to see what you have changed. But I will have a look anyway.

Magalex2x14:
> I propose that we make a new PR, based on the current master and submit your changes in there again.
OK. I will do.

PR #142, opened 2020-11-04
Restructuring of configuration parameters

Ernst79:
This PR aims to change the configuration parameters, as described in #141. 

This final format will be similar to this. This means that this is a breaking change 

```
sensor:
  - platform: mitemp_bt
     discovery: False
     devices:
     - mac: 'A4:C1:38:2F:86:6C'
       encryption_key: '217C568CF5D22808DA20181502D84C1B'
       name: Livingroom
     - mac: C4:7C:8D:6B:4F:F3
       encryption_key: 217C568CF5D22808DA20181502D84C1B
       temperature_unit: F
     - mac: 'B4:7C:8D:6B:4F:D3'
       temperature_unit: 'F'
```

Todo-list
- [x] Add new configuration parameters
- [x] Move `sensor_fahrenheit `to sensor --> `mac` and `temperature_unit`
- [x] Move `sensor_names `to sensor --> `mac` and `name`
- [x] Move `whitelist `to use sensor --> `mac` and `discovery`
- [x] Move `encryptors `to sensor --> `mac` and `encryption_key`
- [x] Remove old parameters
- [x] Code cleaning
- [x] Rewrite readme
- [x] Add faq how to convert configuration. 

Later, additional parameters can be added to sensor level, like `median `and `mean`.

Issue #141, opened 2020-11-02
Changing configuration parameter to sensor level


Ernst79:
Having an idea about the configuration parameters, that I would like to discuss. We now have configuration parameters as lists and dictionaries. In each list we have to define a MAC address, like this:

```
sensor:
  - platform: mitemp_bt
    encryptors:
      'A4:C1:38:2F:86:6C': '217C568CF5D22808DA20181502D84C1B'
    sensor_fahrenheit:
      - 'C4:7C:8D:6B:4F:F3'
    sensor_names:
      'A4:C1:38:2F:86:6C': 'Livingroom'
    whitelist:
     - 'B4:7C:8D:6B:4F:D3'
```

I think it would make more sense to define this per sensor, something like this.

```
sensor:
  - platform: mitemp_bt
     automatic_add: False
     entities:
     - mac: 'A4:C1:38:2F:86:6C'
       encryption_key: '217C568CF5D22808DA20181502D84C1B'
       name: 'Livingroom'
     - mac: 'C4:7C:8D:6B:4F:F3'
       unit_of_measurement: fahrenheit
     - mac: 'B4:7C:8D:6B:4F:D3' 
```

I understand that this is a significant breaking change, but it has the advantage that we can expand this in a second step to define other parameters on sensor level, like `rounding`, `median`, etc.

I’m not saying that we have to make this change, but I see some advantages, despite the breaking change (and quite some work to make this change). Let’s first discuss @Magalex2x14 what you think of this idea and see where we end up.

Magalex2x14:
When I tested the changes with the whitelist today, I had the thought that something needs to be done with these dicts/lists, but I never got to the point of trying to formulate what exactly to do with them)

Looks great to me. We prepare lists for the parser only once at the start of the system, so the complication of this stage will not affect the overall performance of the component. Yes, this is a breaking update, but I think users will support it as the new format is much clearer.

The only thing is, instead of entities, I would write sensors (or devices?). And I also want to change automatic_add to something more sonorous ) If I understand correctly, this is a reference to whitelist: true. Something like autodiscovery, maybe )

In addition, if my memory serves me, there was a request from users to individually configure the averaging and rounding parameters for each device...

Ernst79:
I will have a look tomorrow what the usual parameter naming is. I’ve seen sensors, as well as entities, I thought. I used automatic_add, as it was a parameter for rfxtrx component. autodiscovery sound better, I agree. I will check the developer docs of home assistant. 

[...]

PR #150, opened 2020-11-12
Integration level

### LINKED ISSUE #146: Move to integration domain key
According to [this ADR 0007](https://github.com/home-assistant/architecture/blob/413e3cb248cf8dca766c0280997f3b516e23fb6d/adr/0007-integration-config-yaml-structure.md), we should move mitemp_bt to the integration domain key. 

> New integrations have to put the integration configuration section under the integration domain key if configuration YAML is used.

If I understand it correctly, we should move to the following. 

```
# Allowed
mitemp_bt:
  rounding: True
  etc.

# Not allowed
sensor:
  - platform: mitemp_bt
    rounding: True
    etc.
```

It's not a big deal if we don't do this, but if we want to follow the guidelines, I think we should change this some day.

Ernst79:
This PR 
- changes the name of the component from `mitemp_bt` to `ble_monitor` (breaking change)
- changes the sensor platform to a integration (breaking change)
- includes the changes of beta 0.8.0 (see #142) (breaking change)

Fixes #146 and #141


Magalex2x14 (approved):
Все выглядит хорошо.

Magalex2x14:
Lets make a beta release?

Ernst79:
Shall we first change the repo name?

Magalex2x14:
So I think too ... If we do this now, the current version will stop working (links will break).
But what happens if we roll out the beta without renaming the repository?

Magalex2x14:
Hmm. Look at this: https://stackoverflow.com/questions/5751585/how-do-i-rename-a-repository-on-github/16602311#16602311

Magalex2x14:
It turns out that we can safely rename, github will automatically redirect to a new name.

Magalex2x14:
Let's rename it 🤞🙈

Ernst79:
Ok, I will do it right now. fingers crossed

Ernst79:
Done!

Magalex2x14:
It looks like everything is in order )

Ernst79:
Yep, old links are redirected. Great, gives us some time to fix links

Magalex2x14:
I changed "about" to "Xiaomi Mijia BLE MiBeacon monitor"
image

Magalex2x14:
I tried to reinstall 0.8.0 beta - all OK (folder name stayed as mitemp_bt)

Ernst79:
Ok, great. That is how it should be.

Ernst79:
0.8.1-beta is released, lets try

Ernst79:
image
 It's going too smooth

Ernst79:
image

Also looking good, but these links never work in HACS

Ernst79:
crap, it says (very shortly) "no content to download"

Magalex2x14:
I'm fine, there is no such error.

Ernst79:
has it removed the old mitemp_bt folder and installed it in the new folder?

Magalex2x14:
Yes. All I did was change the configuration to ble_monitor

Ernst79:
Nevermind, it worked on my end as well. Folder has changed as well. 

That was rather easy.

Magalex2x14:
But a little scary :D

Ernst79:
Yes, I agree. 

Has it created new sensors on your end? I got sensors with _2 at the end, but that happened earlier during testing.

Magalex2x14:
Yes. I think we need to add a recommendation that it is worth setting names for each sensor, so that it will be easier to find them later.

Ernst79:
Hmm, if that happens, we might get rid of the mi_ as well

Magalex2x14:
Good idea, yes.

Magalex2x14:
And move the 0.8.1 beta tag to the commit that includes this change (I've done this before)

Ernst79:
Done, How do I move the 0.8.1-beta tag?

Magalex2x14:
I am doing this with SourceTree app, dont know how to do it via website...

Ernst79:
Can you do it than, with adding this note?

Note that your sensor names are most likely also renamed. Look for sensors that start for ble_ (e.g. `ble_temperature_livingroom`). We recommend to use the new [name](#name) option to easily rename and find your sensors.

Magalex2x14:
Yes, I will.

Magalex2x14:
Done.

Magalex2x14:
My entity id's stays the same

image

I will try to disable the component, delete old entities, and enable the component back.

Ernst79:
for me too, only friendly name has changed.

Magalex2x14:
I suspect that when the integration name changes, users will have ble_ id's instead of _2

Ernst79:
I know why. The entity ID is created based on the friendly name on intial setup. The entity ID is linked to (but not the same as) the unique_id. My update doesn't update the unique_id, however, it is updated when moving from sensor.mitemp_bt to ble_monitor. So, for other users, it will update to the new name after updating. Just not for us.

Magalex2x14:
I agree, yes. Everything comes to order after deleting the entities.

Ernst79:
Excellent. when do you want the final release?

Magalex2x14:
Maybe in a couple of days? I will make an announcement on the forum a little later today.

Ernst79:
Oh, I remember one more thing (sorry). In these two lines, there is a double `__`. Can this be changed to one `_`?

`self._unique_id = "cn__" + self._sensor_name`
`self._unique_id = "batt__" + self._sensor_name`

Magalex2x14:
Yes, let's do. I'll move the tag again )

Ernst79:
Done, better do it now.

Magalex2x14:
Tag moved.

Magalex2x14:
This branch does not need to be deleted when we merge - the links to the transition guide point to this branch.

Ernst79:
We can also move the transition guide to the master branch and update the links. Or not?

Magalex2x14:
Yes, let's do it! This does not break anything, and I will add a link to the master in the announcement.

Ernst79:
Ok, will move it in a minute.

Magalex2x14:
Look also at the announcement text. Have I forgotten something?

Hi to all!

A major component update is coming. Serious not by the number of changes, but in their essence.
We collected all the planned breaking changes, and decided to make them together to hurt our users only once :) 

First, we changed the name of the component to __ble_monitor__ (the repository name and the folder name with our component will change accordingly). This had to be done since the old name no longer reflects the essence of the component, and the new name gives us the freedom to implement support for other protocols (in addition to Xiaomi Mijia BLE MiBeacon) and expand the functionality of the component beyond receiving readings from sensors.

The prefix of the sensor entity names will be changed accordingly: from **`mi`** to **`ble`**.

The configuration of the component has been moved from the sensor platform level to the integration level - this also gives us more freedom in the further development of the component's functionality.

The structure of the configuration itself has also changed - the settings are grouped into the "per device" order.

The process of applying changes started today by renaming the repository and publishing version [0.8.1-beta](https://github.com/custom-components/ble_monitor/releases/tag/0.8.1-beta). Let's test its several days, and we will make a full release.

@Ernst prepared a great [guide for configuration updating](https://github.com/custom-components/ble_monitor/blob/integration_level/update_instructions.md). Please read it carefully before updating!
We strongly recommend that you register custom names for each of your sensors before updating (with the `name` option) - it will be much easier to find new entities of your sensors after updating. Do not forget to also delete old entities after updating.

We hope this transition goes smoothly for you, and we will be happy to answer any questions you may have...

Ernst79:
Excellent.

Magalex2x14 (approved):

Magalex2x14:
Everything seems to be ready? )

Ernst79:
Yes, I think it is ready now. Thanks

Magalex2x14:
And thank you ) Let's see how it goes )

PR #154, opened 2020-11-14
Typo fix

dvv:
`sensor.mi_llluminance_foo` -> `sensor.mi_illuminance_foo`


Ernst79:
thanks. This is already fixed in the upcoming 0.8.1 release. See #150 , last commit. But thanks for reporting, much appreciated.

PR #155, opened 2020-11-14
Mark thread as "daemon"

dvv:
Mark thread as "daemon" to help home-assistant better reload

### DIFF
--- a/custom_components/mitemp_bt/sensor.py
+++ b/custom_components/mitemp_bt/sensor.py
@@ -120,7 +120,7 @@ class HCIdump(Thread):
 
     def __init__(self, dumplist, interface=0, active=0):
         """Initiate HCIdump thread."""
-        Thread.__init__(self)
+        Thread.__init__(self, daemon=True)
         _LOGGER.debug("HCIdump thread: Init")
         self._interface = interface
         self._active = active


Ernst79:
Thanks, will leave this for @Magalex2x14 to review.

[Link for explanation docs](https://docs.python.org/3/library/threading.html#threading.Thread.daemon)

Magalex2x14:
As far as I know, a daemon differs from a regular thread in that it terminates automatically when the main thread terminates.
I agree that this can be an advantage for many tasks. However, when I was implementing multithreading in the component, I decided not to use the daemon, because the join method closes the connection with the hci-interface before actually terminating the thread, and I have no idea if there are any negative effects or not in cases where the thread will be terminated before closing the connection. I think this is a reason for research (active connections can be viewed using the btmon utility). @dvv, can you say something about this?

dvv:
@Magalex2x14 I just tend to rely on the system that is believed to correctly releases resources. I've been seeing no negative effects so far -- it just keeps correctly receiving and parsing packets.
However, if we do care, may be consider to receive hci-packets via (say) mqtt subscription and provide a simple command line hci->mqtt relaying daemon to the host-side?

Magalex2x14:
I have something similar in my head (while we have too many tasks that need to be solved before) along with a request for a stand-alone utility/library.

The only negative effect that comes to my mind is a large number of unclosed connections on the hci-interface when the Home Assistant is restarted multiple times. I saw that they remain, but I don't know if they are closed automatically by the system after a timeout, for example.

Tell me, what does "better reload" mean? Have you had any problems restarting HA? And was there any noticeable improvement after your changes?
If not, then I propose to postpone the implementation of this feature for now, but do not forget about it! )

dvv:
It used to get stalled for about 20 seconds on restart
```
homeassistant_1  | Home Assistant attempting to restart.                                                                                                                      
homeassistant_1  | Found 4 non-daemonic threads.
###
HERE
###                                                                                                                              
homeassistant_1  | Restarting Home Assistant                                                                                                                                  
```
which was pretty annoying.

Now it restarts cleanly even without the proposed change -- I start to believe it was because of some other custom component.

So you may consider this PR irrelevant.
Thanks

Magalex2x14:
Hmm... Our component can spawn 4 threads only if data is collected from four interfaces. Our threads terminate in any case (with a timeout of 10 seconds). So yes, I also tend to think that these are not our threads ...

Anyway, thanks for the reminder and your participation!

PR #159, opened 2020-11-17
Realtime

Magalex2x14:
Changes:

- the state of the binary sensors is updated in real-time
- initial support for MCCGQ02HL sensor #115
- included a modified version of aioblescan lib to workaround Extended Advertisements feature problem #61
- other minor changes


Magalex2x14:
We can roll out a new beta for testing MCCGQ02HL after you review

Ernst79:
I will review it this evening

Ernst79:
Can you check the README, on the refresh rate of the sensor (realtime)? Perhaps different for the battery

Magalex2x14:
I propose to write something like "beta test, data to be confirmed" for now - I'm not sure about the information we have.

Ernst79:
Added the MAC_REGEX check for the mac parameter. Was apparently missing or lost somewhere

Magalex2x14:
I will make a new beta - let users test it

Ernst79:
Perfect. I will check the last file tonight. So far, not problems found. I only have to check sensors.py

Ernst79:
- Fixed some pydocstyle errors
- In the last part of sensor.py, I've added the `self._device_class` for 3 binary sensors. 
- You also had a `self._measurement` twice, for the same sensor. 

I don't have these binary sensors myself, so could not test these in real life.

Magalex2x14:
Thank you! I will update the 0.8.3-beta tag

Ernst79:
Tested, works fine as well.

andrewjswan:
> On my end, it is all working fine for the normal TH sensors, they show up within the set period.

I have 8 thermometers (LYWSD03MMC), one alarm clock (CGD1) and one opening sensor (MCCGQ02HL). After installing 0.8.3.Beta after rebooting the XA, not all devices started working, 3 did not appear in the XA even after an hour of waiting. I returned to 0.8.2.Beta and after 5-10 minutes I had all the devices and all sensors in working order.

Ernst79:
@Magalex2x14 I've figured out the restore state functionality upon a restart. If you want, I can add it to the realtime branch. Other option is to wait till the next release. Let me know what you want. 

@andrewjswan I think it would be helpful for @Magalex2x14 if you can try again and provide your logging if it doesn't work. For me, it is working fine, so it is hard to tell what is going on on your end.

Magalex2x14:
@Ernst79 I think it can be done in this branch, but is it possible to have the option to disable restore state functionality? For example, the option "restore_state: True" by default.

@andrewjswan As Ernst said, logs are needed. Also, see if there is any dependency on the `discovery` option. Try to change it. Are all of your sensors included in the configuration?

Ernst79:
Yes, sure. Will add that tomorrow.

andrewjswan:
I dont see any errors in log, and in 0.7.X and in 0.8.2 all ok, but in 0.8.3 not work...

andrewjswan:
Yes all my sensor included and worked in 0.8.2 without problem (except open sensor)... And all sensor have name ...

Ernst79:
Please enable debug logging and show the messages. Even if there are no errors.

```
logger:
  default: warn
  logs:
    custom_components.ble_monitor: debug
```

Ernst79:
I've added the restore state upon restart functionality. Will add documentation tomorrow, but I used your suggestion `restore_state: True`. I can't test it for binary sensors, do you have such a device? I don't expect much issues however.

Magalex2x14:
I have some news. Good and bad :)
1. Good. I bought myself a MCCGQ02HL today.
2. Bad: Xiaomi (at the end of July, as I was able to find out) enabled the encryption of the request body in which bind_key is passed. That is, all of our instructions for extracting an encryption key from intercepted traffic are no longer valid.
3. Good, but not very good) I figured out how to get the key from traffic, but this operation has become much more complicated...
4. Good on the one hand, but bad on the other. I caught all the errors that users posted about 🤣

```
2020-11-19 01:36:38 ERROR (MainThread) [homeassistant.helpers.entity] Update for sensor.ble_opening_e4aaec34b657 fails
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 278, in async_update_ha_state
    await self.async_device_update()
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 474, in async_device_update
    raise exc
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/ble_monitor/sensor.py", line 1076, in update
    self._device_state_attributes["rssi"] = round(sts.mean(self.rssi_values))
  File "/usr/local/lib/python3.8/statistics.py", line 315, in mean
    raise StatisticsError('mean requires at least one data point')
statistics.StatisticsError: mean requires at least one data point
2020-11-19 01:36:40 ERROR (MainThread) [homeassistant.helpers.entity] Update for sensor.ble_light_e4aaec34b657 fails
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 278, in async_update_ha_state
    await self.async_device_update()
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 474, in async_device_update
    raise exc
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/ble_monitor/sensor.py", line 1025, in update
    self._device_state_attributes["rssi"] = round(sts.mean(self.rssi_values))
  File "/usr/local/lib/python3.8/statistics.py", line 315, in mean
    raise StatisticsError('mean requires at least one data point')
statistics.StatisticsError: mean requires at least one data point
2020-11-19 01:36:45 ERROR (MainThread) [homeassistant.helpers.entity] Update for sensor.ble_opening_e4aaec34b657 fails
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 278, in async_update_ha_state
    await self.async_device_update()
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 474, in async_device_update
    raise exc
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/ble_monitor/sensor.py", line 1076, in update
    self._device_state_attributes["rssi"] = round(sts.mean(self.rssi_values))
  File "/usr/local/lib/python3.8/statistics.py", line 315, in mean
    raise StatisticsError('mean requires at least one data point')
statistics.StatisticsError: mean requires at least one data point
2020-11-19 01:36:47 ERROR (MainThread) [homeassistant.helpers.entity] Update for sensor.ble_light_e4aaec34b657 fails
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 278, in async_update_ha_state
    await self.async_device_update()
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 474, in async_device_update
    raise exc
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/ble_monitor/sensor.py", line 1025, in update
    self._device_state_attributes["rssi"] = round(sts.mean(self.rssi_values))
  File "/usr/local/lib/python3.8/statistics.py", line 315, in mean
    raise StatisticsError('mean requires at least one data point')
statistics.StatisticsError: mean requires at least one data point
```

Magalex2x14:
2020-11-19 02:01:44 ERROR (Thread-5) [root] Uncaught thread exception
  File "/usr/local/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/homeassistant/.homeassistant/custom_components/ble_monitor/sensor.py", line 442, in run
    sensors[op_i].schedule_update_ha_state(True)
  File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 422, in schedule_update_ha_state
    assert self.hass is not None
AssertionError

Magalex2x14:
Quick note:
https://developers.home-assistant.io/docs/core/entity/

> > A comment on discord about this issue points the finger to a race condition:
> > > The platform implements the Entity interface incorrectly, using it before the entity has been added to home assistant.
> > 
> > 
> > Simone
> 
> Some more details for @hakana:
> 
> > Make sure to not use the Entity interface before the entity has been added to home assistant, it will work. That's how it should be used.
> > In particular, do not update during setup.
> 
> Maybe an update message came in during setup and this created the race condition.
> 
> Suggestion from dev is to "Subscribe in async_added_to_hass() https://developers.home-assistant.io/docs/core/entity/#async_added_to_hass"
> 
> Simone

It looks like we are trying to update the binary sensor too quickly. In the case of the measuring sensors, this does not happen due to the "period" delay before the first update...

Ernst79:
I have added the async_added_to_hass() already, yesterday. It was needed for restore_state functionality as well, so it should be in there now. Did you get this error with my latest additions?

andrewjswan:
> Please enable debug logging and show the messages. Even if there are no errors.

Ok, i eneble debug for this integration, install 0.8.3 Beta, reboot and look, after 10-15 minutes i send log by email to @Magalex2x14 :)

andrewjswan:
@Ernst79 @Magalex2x14 With debugging enabled in the log, all sensors (except for the door opening sensor) are working. 
All data is available. 0.8.3.Beta
Just turned off the debugging mode and reloaded the HA. I will see how it will be this time. 
The opening sensor is most likely not pulled up because there is no one at home and there are no events from it, although in the gateway all the data is there and the information is transmitted to the gateway (the log is visible, the current state is the same).

andrewjswan:
> 2\. Bad: Xiaomi (at the end of July, as I was able to find out) enabled the encryption of the request body in which bind_key is passed. That is, all of our instructions for extracting an encryption key from intercepted traffic are no longer valid.

Vevs mod not solve this problem?

Magalex2x14:
@Ernst79 no, I haven't tried with your changes. Now I'll look at the code and try.

@andrewjswan enabling the debug logging does not affect the operation of the component in any way, it sends messages to the logger anyway, setting the logging level works only as a filter. There are no checks and dependencies on the level of logging in the component.

The mod works on a different principle, and intercepts api requests before encryption, so this did not affect its work in any way. But not everyone has an android :) Later I we will create a separate issue about methods for obtaining keys.

andrewjswan:
> enabling the debug logging does not affect the operation of the component in any way, it sends messages to the logger anyway, setting the logging level works only as a filter. There are no checks and dependencies on the level of logging in the component.

Yes, I understand this, but I can see it the first time I tried it, or something wrong or some other reason. Now I've turned off the debug and everything works as well as before.

Magalex2x14:
It seems I have to fixed it. We need to test. 
Good news for @andrewjswan  - door close timeout events are also coming from the sensor, I added the ext_state attribute, which becomes =2 on timeout. The status of the binary sensor entity is updated almost instantly (the light sensor has a slight delay, but it is due to the operation of the sensor itself).

@ Ernst79 Regarding restore_state - it doesn't seem to work for binary entities, and sometimes I get errors that I can't explain yet (I didn't save the log, I was doing something else)... Tomorrow I'll look further.
In addition, there is a subtle point with restore_state: if during HA reboot (or shutdown time) the sensor changed state, then HA will have an incorrect status until a new event from the sensor... It seems that the MCCGQ02HL sends messages only upon the occurrence of an event (physically), and does not repeat its status periodically...

andrewjswan:
Good news, how about RSSI for MCCGQ02HL i dont see any RSSI value in 0.8.3 beta

Magalex2x14:
@Ernst79 I made the option restore_state = False by default (for debugging period), now I will update the tag for 0.8.3-beta

Ernst79:
> then HA will have an incorrect status until a new event from the sensor...

We can’t do anything about that unfortunately, but it will update to the correct state after the period. 

If we can’t get the restore state working, we can always remove it for the binary sensors for now.

Magalex2x14:
Yes, I understand it.
I'm done for today, it seems. Tomorrow I'll take a closer look at this, and we'll make a decision.

Ernst79:
All working ok for the normal sensors.

custom_components/ble_monitor/sensor.py
```diff
@@ -1027,10 +1026,11 @@ async def async_added_to_hass(self):
             self.ready_for_update = True
             return
         old_state = await self.async_get_last_state()
+        _LOGGER.info(old_state)
         if not old_state:
             self.ready_for_update = True
             return
-        self._state = old_state.state
+        self._state = True if old_state.state == "on" else False
```
Ernst79, line 1033: I propose to change this to `self._state = True if old_state.state == STATE_ON else False`

Magalex2x14: Yes, of course! I keep forgetting about these constants...

Ernst79:
Just a (very) minor detail, but I finally figured out how to do comments in  the code directly :-)

Magalex2x14:
@Ernst79 It seems that I have solved all the current problems, including those with restoring the state of binary sensors.
However, I think it is worth discussing the default state of this option.
If enabled by default:
- Acceptable for measuring sensors, but still controversial... For example, after a reboot, I clearly distinguished between the statuses of unavailable and unknown, and this helped to diagnose problems with reception. In principle, the issue is solved by additional automation that checks the entity's last_updated attribute, but personally, I'm still uncomfortable ...
- For binary sensors, during the host reboot, the state may change, and this may be "unsafe", since important automation (movement, the opening of windows and doors) is usually tied to binary sensors.

My thoughts:
1. Make the option disabled by default and add a warning to the documentation so that the user makes a decision consciously.
2. point 1 + make this option "per device" since this makes it possible to use different behavior for measuring and binary sensors, for example.

Ernst79:
@Magalex2x14 I agree with you that for binary sensors, this can be tricky. So lets keep it disabled by default. I'll update the docs accordingly. 

Regarding making this a "per device" setting, I think that we have a lot more options that can also be a "per device" setting (like median, rounding, etc). I would propose that we keep it as it is for now and make a final release. In a future update, we can look into moving options to a "per device" option, but lets not try to do it all in one update. 

Talking about the per device options, I would suggest that we keep a setting at integration level, that is overruled by a "per device" setting.

Magalex2x14:
OK. Let's do it. Do you think we're ready for release?

Magalex2x14:
By the way, the issue of device classes was mentioned earlier, and I am also confused by the fact that the prescribed class does not affect the sensor behavior in any way. As far as I understand, the appropriate icon and text representation of the state should be set. But this is not the case with our binary sensors. Yesterday I tried to search in this direction, but did not find anything... In the case of measuring sensors, everything works. Do you have any thoughts on this?

andrewjswan:
I looked at the properties of the sensors, the temperature sensors have only class, the icon is visible from the class, the humidity sensors have both class and icon, the battery has only class and unit of measurement. Maybe for the sensors of opening and light, you should also set the icon (although there is obviously need to set two, for each status).

andrewjswan:
But i dont see DEVICE_CLASS_OPENING in source code - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L12-L24
Maybe need add it, and use like Temperature? https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L829

Ernst79:
> But i dont see DEVICE_CLASS_OPENING in source code - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L12-L24
> Maybe need add it, and use like Temperature? https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L829

I didn't find DEVICE_CLASS_OPENING in [const.py](https://github.com/home-assistant/core/blob/dev/homeassistant/const.py)

andrewjswan:
The some for DEVICE_CLASS_LIGHT and maybe for other to:
**DEVICE_CLASS_SWITCH** - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1128
**DEVICE_CLASS_LIGHT** - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1141
**DEVICE_CLASS_OPENING** - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1155

Ernst79:
@andrewjswan But the DEVICE_CLASS_# should be defined somewhere. I normally look them up in the const.py file from Home Assistant core. But I can't find the ones you are referring to. 

The [binary sensor integration](https://www.home-assistant.io/integrations/binary_sensor/) does however show the available device types

andrewjswan:
> I didn't find DEVICE_CLASS_OPENING

Look https://github.com/home-assistant/core/blob/master/homeassistant/components/binary_sensor/__init__.py#L97-L122
And https://github.com/home-assistant/core/blob/master/homeassistant/components/xiaomi_aqara/binary_sensor.py#L4-L8

Ernst79:
Aha, there they are. Thanks

andrewjswan:
**DEVICE_CLASS_SWITCH**  - https://github.com/home-assistant/core/blob/master/homeassistant/components/switch/__init__.py#L39

Ernst79:
I think switch is not a correct type, is it? A switch is something you can put on, but In this case, it only shows if it is on or off. We have to pick one from the binary sensor class.

Magalex2x14:
If you look at the contents of these constants, you will see that these are the same strings themselves. I tried it yesterday - it doesn't change anything. But it is worth using the constants, yes.

Magalex2x14:
If you wish, do it yourself.

Ernst79:
I would propose to use DEVICE_CLASS_POWER in stead of Switch. Switch is not a binary sensor type

andrewjswan:
> I tried it yesterday - it doesn't change anything.

Maybe this interferes - https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1150
https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1124
https://github.com/custom-components/ble_monitor/blob/realtime/custom_components/ble_monitor/sensor.py#L1137

measurement for binary sensor strange property .... :) i dont see measurement in xiaomi_aqara/binary_sensor - https://github.com/home-assistant/core/blob/master/homeassistant/components/xiaomi_aqara/binary_sensor.py#L294-L308

Magalex2x14:
The entity has standard inherited properties. This property is not included in their number. I am free to assign any properties that I need.

For the purity of the experiment, you can create a test entity at the start of the component and look at its behavior.

Ernst79:
Changed, I used DEVICE_CLASS_POWER for the switch of the Mosquito Repellent sensor. I also added `BinarySensorEntity`, to the SwitchingSensor class, perhaps that is needed to being recognized as a binary sensor instead of a normal sensor. 

Please try it, I don't have a binary sensor here

Magalex2x14:
Unfortunately, this did not solve the problem. I tried deleting entities with the component disabled - it didn't help.
image

Magalex2x14:
I'm thinking of trying to ask a question on discord or create issue in the HA core repo.

Ernst79:
It should NOT be sensor.ble_###, but binary_sensor.ble_###. Don't we need to move the entire thing to binary_sensor.py?

Magalex2x14:
Very sharp eye! I think it's worth a try. We can also try to specify the domain somehow.

andrewjswan:
> It should NOT be `sensor.ble_###`, but `binary_sensor.ble_###`

It makes sense! Excellent work!

Ernst79:
It is now all loaded as a sensor platform, but the binary sensors should be in the binary_sensor platform. 

This is a big change: it would mean the connection has to move to __init__,py measurement, sensors in sensor.py, binary sensors in binary_sensor.py.

Magalex2x14:
Yes, and I think it's worth doing anyway. I also think that the release should be done before that.
But we always loaded as a sensor platform and it worked earlier for a repellent if my memory serves me right. I'll go and see the WX08ZM issue.

andrewjswan:
> We can also try to specify the domain somehow

Like - https://github.com/home-assistant/core/blob/master/homeassistant/components/smarty/__init__.py#L57-L58

Ernst79:
Yes, that is how it should be. But you don't want the BLE processing twice, so that should move to `__init__.py`, I think. 

We can keep it as `sensor` for now, release it, and change it in a next release?  It would mean a breaking change for the binary_sensor users, but in the meantime. people can use the sensor with on/off. We can add a note in the release notes that we are working on it for a next release.

andrewjswan:
> This is a big change: it would mean

I think it should be done now. Then, it will be more difficult, flood control devices SJWS01LM have entered the market, perhaps soon there will be more. And there are still few users of new sensors. :)

andrewjswan:
> It would mean a breaking change for the binary_sensor users

There are not so many of us, and so far everybody is testing and ready to change as there is no release. Then, when the running sensor is automated, interconnected and output to the GUI, the changes will be more complex. I would prefer to have binary sensor support in the release so that I can build on it without worrying.

Ernst79:
Yep, I think I have to agree on you with that. If people want, they can always use the beta.

Magalex2x14:
I agree. Nobody drives us. The only ones I really want a release for are Intel NUC users.
We just have to try. If it turns out that this is too complicated, and entails a lot of new problems, let's get back to talking about the "intermediate" release again.

andrewjswan:
This is a great solution, I don't know Python, but it seems to me that if everything is already working, then divide from one module into 3, there shouldn't be a big problem and I don't think it should cause any problems. But life is a strange thing. But I hope for the best! Thank you so much!

Magalex2x14:
In fact, there are many subtle points (for example, the `add_entities` method becomes available only for the platform, but we are going to add entities from the integration level). But so far I do not see any insurmountable problems.
I'll try it today / tomorrow.

Magalex2x14:
Last night I failed in my attempts to resolve the issue of platform separation head-on.

Here's what I found out: 

1. There is no way to get and use the `add_entities` method at the component level.
2. Adding and updating entities should take place at the platform level.
3. The key to the "correct" solution to the problem is the use of [dispatchers](http://dev-docs.home-assistant.io/en/master/api/helpers.html#module-homeassistant.helpers.dispatcher), as I understand it. But this will require a major change in the architecture of the component, and I do not yet know how to do this, and how long it will take.

I see an alternative solution in the following:

1. The HCIdump thread is initiated and works at the integration level, and splits the incoming messages into two separate queues (one for each platform).
2. Each of the platforms starts its own separate thread that parses the corresponding message queue and creates/updates the entities of the corresponding sensors.

- An alternative approach will still require a revision of the component architecture.
- Another point requiring attention will be that now our scanner is restarted every 60 seconds (this increases the stability of receiving data, because if the scan stops due to another component or interface errors, it will be automatically restarted. Thus, it will be necessary to come up with a way to control the HCIdump thread from the platform, or to catch the scan stop events (not enough knowledge so far). This problem still occurs with the "correct" method above.
- An alternative approach will further increase the CPU load (I can't say how much, but I think it is noticeable.)

In general, I am still leaning towards the option "leave it as is", because in any case, the path to solving the discussed problem will be too long for us, there are not enough competencies, it will take a lot to study and try a lot.
Since both options require serious work, we should strive for the "correct" solution, so as not to do the work twice.

Ernst79:
I agree on your analyses, I also tried it yesterday, but also failed on the `add_entities`. Unfortunately, my knowledge is also limited how to solve this. 

What is your proposal now, make a release now and take our the time and try to solve it in the future?

Magalex2x14:
Yes, I think we need to make a release and open a new issue for this problem.

Ernst79:
At this moment, I thing your first alternative solution looks the most promising. 

Lets make a final release. We only have to add the release notes in info.md. Perhaps add a note about the binary_sensors.

Magalex2x14:
Yes, I agree. Do you want to do this?

In fact, everything goes to the fact that we are moving towards the asynchrony of the component (which corresponds to the architecture of the HA core) - we also need to dive into this (async magic), and it is quite possible that this will also automatically solve the problem with platform separation.

Ernst79:
Yes, ok. I'll make the release this afternoon

Magalex2x14:
Let me summarize the changes since 0.8.1 (from a user perspective):
- support for Extended Advertising features (Intel NUC and other BLE5 adapters, solves #61)
- `restore_state` option
- processing of events from binary sensors in real-time
- support for MCCGQ02HL
- sensor entities are created when **any** device activity is detected (almost instantaneous status change from unavailable to unknown when the device is working and is in the reception area)

My observations about MCCGQ02HL:
- the sensor sends messages only at the moment of an event occurrence, does not notify about its state periodically
- there is information about the battery, but very rarely (only with actual level change?). Left it at open state at night, and in the morning the battery entity last_update attribute was "6 hours ago".

PR #168, opened 2020-11-22
[WIP] Config flow for ble monitor

koying:
Besides the obvious niceness of being able to configure the integration via the webui, I wanted to have the notion of device added, so I started working on the config flow.

Current version works fine. Only thing is that we cannot configure devices via the webui, but I don't even know if it's possible in HA, currently.
The yaml configuration is picked up, so no changes to be done to the config.

image
image
image


Ernst79:
Great work, thanks. I think this is a great improvement.

I see that it is still a WIP, can you tell us what you what you want to add/change?

Do you want us (@Magalex2x14 and @Ernst79) to start reviewing, or shall we wait till you think it is finished?

custom_components/ble_monitor/config_flow.py
```diff
@@ -0,0 +1,157 @@
...
+DOMAIN_SCHEMA = vol.Schema(
+    {
+        vol.Optional(CONF_ROUNDING, default=DEFAULT_ROUNDING): cv.boolean,
+        vol.Optional(CONF_DECIMALS, default=DEFAULT_DECIMALS): cv.positive_int,
+        vol.Optional(CONF_PERIOD, default=DEFAULT_PERIOD): cv.positive_int,
+        vol.Optional(CONF_LOG_SPIKES, default=DEFAULT_LOG_SPIKES): cv.boolean,
+        vol.Optional(CONF_USE_MEDIAN, default=DEFAULT_USE_MEDIAN): cv.boolean,
+        vol.Optional(CONF_ACTIVE_SCAN, default=DEFAULT_ACTIVE_SCAN): cv.boolean,
+        # vol.Optional(
+        #     CONF_HCI_INTERFACE, default=[DEFAULT_HCI_INTERFACE]
```
Ernst79, line 62: I think if a list doesn't work, you can use:

```
vol.Optional(CONF_HCI_INTERFACE, default=DEFAULT_HCI_INTERFACE): cv.positive_int,
```

This will allow only one HCI interface to be setup, but that will be sufficient for most people. It at least gives users an option

koying: > Unfortunately, I got

That's why WIP ;)
I probably didn't test all the cases.

Do you have something configured in yaml ?

koying: > I think if a list doesn't work, you can use

Yep. Good point. Will do.
I'll also look for device list, but I have little hope

Ernst79: I tun 0.118.2. I removed config, restarted and added via UI. Then removed it, added again, but somewhere, it got messed up. Now I have the above error after each restart. Setup failed for ble_monitor: No setup function defined.

Ernst79:
The rfxtrx integration has a possibility to set configuration parameters for devices. Perhaps we can use something similar?

Magalex2x14:
Hey! Great start! I was stopped from implementing config flow (or rather, from working with devices) by the fact that everything is built around asynchrony (as far as I understand), and I still don't understand much about this. I will definitely look at this, but a little later.

> Are sensors that require a bind key detected without entering a bind key (e.g. by mac address). I know that they won't work without bind key, but if we can add the mac or sensor_name as a device/non working sensor, we can make it show up in the device option menu, where the user can add a bind_key. In theory, this should work.

Yes, sensors regularly send packets with no payload. It is thanks to this that we are now creating entities before the arrival of useful data (see the code section in the parser with a comment `# check data is present`). The key is only needed to decrypt the payload (which may not be presented).

koying:
> The rfxtrx integration has a possibility to set configuration parameters for devices. Perhaps we can use something similar? But I realize that the rfxtrx devices are first detected. Only after detection you can add configuration parameters.

Cool. I'll have a look.

Re your issue with "no setup", I cannot reproduce.
Could you reproduce in debug and provide me with the logs? Thx

koying:
> and I still don't understand much about this

Yeah. That thing is a bit of a black box with poor doc, so I mostly go to existing implementations to gather bits and pieces ;)

Ernst79:
> Re your issue with "no setup", I cannot reproduce.

I think I missed something while copy-ing your code. I started over by copying your code and now it works. 

image

Ernst79:
> and I still don't understand much about this

Some nice reading stuff for async, for under the christmas tree

https://realpython.com/async-io-python/

https://developers.home-assistant.io/docs/asyncio_working_with_async/

koying:
Thanks.
I was talking about the config/options flow of HA, but yeah, async itself is a bit hard to grasp

custom_components/ble_monitor/config_flow.py
```diff
@@ -0,0 +1,157 @@
...
+        return await self.async_step_user(user_input)
+
+    @callback
+    def _show_user_form(self, errors=None):
+        return self.async_show_form(
+            step_id="user", data_schema=DOMAIN_SCHEMA, errors=errors or {}
+        )
+
+class BLEMonitorOptionsFlow(config_entries.OptionsFlow):
+    """Handle SpeedTest options."""
```
Ernst79, line 115: Hope you don't mind when I make suggestions during the development. 

Change to
"""Handle BLE Monitor options."""

Ernst79:
I've created a small PR on your branch which adds
- manufacturer info to the device
- device info for binary sensors

Magalex2x14:
Wow) It remains to figure out how it works :)

Ernst79:
@Magalex2x14 It isn’t as complicated as the stuff you do with this bluetooth monitoring. 

@koying I’ve created a new PR which adds multiple HCI inputs. But I think it would be good if we create a branch here where we can work together on this. I can give you the necessary rights.

koying:
@Ernst79 Sure, no problem for a branch here.
I'm trying to figure out a way to include devices. The rfxtrx integration is a good starting point, just trying to figure out if it's possible to add/modify a device and return to the main dialog...

Also, we should decide whether yaml or config takes precedence if both are present.
Currently, yaml is ignored as soon as the first restart after the config-enabled version is installed (besides devices, currently, ofc). This is inline with HA standards, but I'm not sure to be ok with that :)

Ernst79:
This is an interesting page for that. https://developers.home-assistant.io/docs/data_entry_flow_index

I thought the tuya integration does also something with a flow for devices.

koying:
Devices seems ok. Only initial creation implemented.
I just posted this interim commit for testing

Ernst79:
I will create a new branch tonight/tomorrow (`config_flow`). I will make it based on the `binary_sensor` branch as this one will be the basis for the next beta and has significant changes by @Magalex2x14. I'll add your changes into the `config_flow` branch and will grant you access when I'm finished. In the meantime, you can continue in your own branch.

Magalex2x14:
Since the configuration in the web interface can be changed on the fly, we need to think about which options require a component/platforms reload and decide how to work around it. I do not yet understand what changes in the interface entail what calls. It may be possible to restart the component every time a change is made and the submit button is pressed.
In addition, unloading entries implies the need to free resources, as far as I understand. In our case, this definitely requires a platform restart.

Ernst79:
Guys, I've opened #173 with the `config_flow` branch. I propose we close this one and continue in #173. I will grant access to @koying. @Magalex2x14 Please copy your remark to this new PR.

Ernst79:
Closing this one, we continue in #173

PR #172, opened 2020-11-26
Binary sensors

### LINKED ISSUE #166: Use binary sensors
**Problem description**
Follow up of discussion in #159

MCCQ02HL and WX08ZM sensors are currently discovered as `sensors` with on/off state. However, these sensors are actually `binary_sensors`, not `sensors`. This causes that the sensors are only able to show `on/off` and not e.g. `closed/open`. 

**(Possible) Solution**
This issue is to discuss how to solve this. The final goal should be that binary sensors are discovered with the proper device class, and show their appropriate state. 

Proposal for device classes
MCCQ02HL --> `open/close` (DEVICE_CLASS: opening)
MCCQ02HL --> `light/dark` (DEVICE_CLASS: light)
WX08ZM  --> `power detected/no power` (DEVICE_CLASS: power)

**To Do**

To achieve this, we should 

- [x] move the monitoring to `__init__.py` (integration level)
- [x] keep the add_enties in `sensor.py` and `binary_sensor.py` at platform level)
- [x] move BLE ADV parser to HCIdump thread because this is the only way we can sort messages (see next task)
- [x] split the incoming messages into two separate queues (one for each platform)?
- [x] move the `SwitchingSensors` to `binary_sensor.py`

**Findings of other examples that could help**
I think the [smarty](https://github.com/home-assistant/core/tree/dev/homeassistant/components/smarty) can be used as an base example, although it has part of the code in a pypi package (pysmarty). 

What I found in the smarty component, is that there are two links between `__init__.py` and e.g. `sensor.py`
Data
- in `__init__.py`, data is collected in `smarty` and this is passed to `hass.data[DOMAIN]`
```
    smarty = Smarty(host=host)
    hass.data[DOMAIN] = {"api": smarty, "name": name}
```
- in `sensor.py` it gets the data via `smarty = hass.data[DOMAIN]["api"]`

Updating
- It uses `dispatcher_send(hass, SIGNAL_UPDATE_SMARTY)` in `__init__.py` after receiving data 
- In `sensor.py` it uses this in the entity. 

```
    async def async_added_to_hass(self):
        """Call to update."""
        async_dispatcher_connect(self.hass, SIGNAL_UPDATE_SMARTY, self._update_callback)

    @callback
    def _update_callback(self):
        """Call update method."""
        self.async_schedule_update_ha_state(True)
```

Not entirely sure how this update link exactly works.

Magalex2x14:
closes #166


Magalex2x14:
I think we can do a public beta.

Magalex2x14:
I checked the operation of binary sensors. And processing the battery data in binary_sensors (I didn't have to wait 24 hours, MCCGQ02HL sends all the data if the battery is removed and inserted back).

Ernst79:
It works fine here, so OK to me to make a beta release

Ernst79 (approved):

Magalex2x14:
Added some fixes. There are important ones among them, so also updated 0.8.4-beta tag.

Ernst79:
Can I merge your changes in config_flow?

Magalex2x14:
Yes, sure.

Ernst79:
Just to make sure we don't get too far behind

Magalex2x14:
I deliberately tackled small things that require long debugging and do not change the code much.

Ernst79:
Sorry, seems I've done something wrong

image

I merged your changes, but it also deleted the binary_sensor branch. Oops

Magalex2x14:
now I'll try to restore

Ernst79:
I compared the two branches and created a PR to merge, but seems that is not the way to do this?

Magalex2x14:
There seems to be a checkmark somewhere, whether to delete the original branch after merging

Ernst79:
Ok, thanks. Learned something again. It is back online.

Ernst79:
Can we merge this and release this as final, or are you still working on this one?

Magalex2x14:
No, no, this work is over. We can release it, yes. The only thing is that there is ambiguity with the kettle, but I think we can leave it as it is.

Ernst79:
Ok, I will make a release.

Magalex2x14:
Thank you

PR #173, opened 2020-11-26
Config flow

Ernst79:
**Description**
This PR will add a config flow to BLE monitor.

This PR is a merge of the work of @koying (thanks!) (#168) and the `binary_sensor` branch, which is the basis of beta 0.8.4. 

**ToDo list**

- [x] Fix import of devices (and device parameters) from YAML
- [x] Decide on whether YAML import or config flow is most important
  - [x] Disable option flow when YAML is used
  - [x] Disable removing and adding integration when YAML is used (if possible)
- [x] Option flow for changing devices
- [x] Check on only allowing one integration in the UI
- [x] add logo (see instructions [here](https://developers.home-assistant.io/blog/2020/05/08/logos-custom-integrations/))
- [x] Add/modify documentation
- [x] "Manually" validate the MAC address and encryption key format

Feel free to add things to the ToDo list

**Tests**
The following tests have been checked by @Ernst79 
YAML mode
- [x] Import of YAML with all options
- [x] Import of YAML with no/partial options
- [x] Change in YAML changes BLE monitor
- [x] Check what happens on a name change in YAML
         _New device is created, old one is still a device, but sensors are `unavailable`. They can be deleted. **Device can be deleted with service `ble_monitor.cleanup_entries`** Instructions to be added in the docs_
- [x] Check on validation of input (mac, encryption key)

Switching from YAML to UI mode
- [x] Check what happens when commenting out YAML
        _All devices and entities still exist, integration parameters are according to setting as per old YAML, settings are available to change in the UI. Excellent!_

UI mode
- [x] Create integration (user input flow)
- [x] Options (options flow)
- [x] Adding devices (add_device flow)
- [x] Modify integration settings
- [x] Modify device settings
- [x] Check on defaults
- [x] Check on validation of input (mac, encryption key)
- [x] Check on delete and adding an integration while using YAML

Switching from UI mode to YAML
- [x] Check what happens when commenting out YAML
        _It loads the YAML config, as expected_

**Ideas for future enhancements **
Not required before a release, but these ideas can be implemented in the future. 
- [x] Show the device name instead of the MAC address in the `devices` drop down menu
- [x] Automagically prefill the `devices` drop down list with discovered devices for easy configuration
- [ ] Add a `delete device` option in the same drop down menu where you can select `Add device`. It should show a drop down form where you can select devices to delete.
- [ ] We could look into the possibility that clicking submit will delete the orphaned/dangled entities and devices.
- [ ] After deleting a device, it isn't possible to remove the sensor entities. A restart is required before this is possible. Unload entities from the registry would probably solve this. 
- [ ] Add notification when a device is detected


Magalex2x14:
Since the configuration in the web interface can be changed on the fly, we need to think about which options require a component/platforms reload and decide how to work around it. I do not yet understand what changes in the interface entail what calls. It may be possible to restart the component every time a change is made and the submit button is pressed.
In addition, unloading entries implies the need to free resources, as far as I understand. In our case, this definitely requires a platform restart.

Magalex2x14:
We can stop the entire integration by calling `blemonitor.shutdown_handler` (available through `hass[DATA]`).
In this case, `None` will be sent to the platforms queues, upon receiving which the platform will also stop, but the thread will not fully terminate, since it is a daemon, and its resources are released only when the main thread ends (if I understand everything correctly).

Therefore, the platform termination procedure must include a call to the `.join()` method, which blocks until the thread is completely unloaded. The problem is that this method cannot be called from the terminating thread itself (deadlock).

I see a way out in registering a handler for a **custom reboot event** (by analogy with `blemonitor.shutdown_handler`) during platform setup. Since the handler is called from the main HA thread, there will be no problem with `.join()` execution.

Magalex2x14:
> Decide on whether YAML import or config flow is most important

I vote for YAML priority. I liked how it was done in HACS. There, when there was configuration in yaml, there was corresponding text in the integration tile and options in the web interface were not available (if I remember correctly).

Also, I recall a conversation with one of the users who thanked for the fact that our integration is configured in yaml )

Ernst79:
I was unsure about the location of the code below in `sensor.py` and `binary_sensor.py`, should it be in the async_entry_setup or async_platform_setup. I've putted it in the entry_setup now, and it seems to work fine, but please check if this is the right location.

```
    blemonitor = hass.data[DOMAIN]
    bleupdater = BLEupdater(blemonitor, add_entities)
    bleupdater.start()
```

Ernst79:
> I vote for YAML priority. I liked how it was done in HACS. There, when there was configuration in yaml, there was corresponding text in the integration tile and options in the web interface were not available (if I remember correctly).
> 
> Also, I recall a conversation with one of the users who thanked for the fact that our integration is configured in yaml )

Seems like a good solution, but this requires the following:
- We need to "disable" the option flow, when YAML config is loaded. 
- Additionally, we should preferably also disable the possibility to remove the integration in the UI and add a new one in the UI, when YAML config is loaded. 

Not sure the second one is possible, but it is worth investigating.

Magalex2x14:
According to [devdocs](https://developers.home-assistant.io/docs/config_entries_index/#setting-up-an-entry):

During startup, Home Assistant first calls the normal component setup, and then call the method async_setup_entry(hass, entry) for each entry. If a new Config Entry is created at runtime, Home Assistant will also call async_setup_entry(hass, entry) (example).

It's not very clear to me what config.entry is in essence. From the documentation it follows that there can be several of them, and there can be several calls to async_entry_setup. If in our case this is excluded, then you can start the integration from async_entry_setup. If not, then we have to think ...

Ernst79:
I've read that part, but it wasn't clear to me either. Just guessing, but perhaps it has to do with the fact that you can add multiple integrations in the UI of the same component. This can be blocked however in the code.

koying:
When a yaml exists for an integration, it is injected in the config flow via the "import" step (https://github.com/custom-components/ble_monitor/blob/config_flow/custom_components/ble_monitor/__init__.py#L88) which created the config_entry.

Due to https://github.com/custom-components/ble_monitor/blob/config_flow/custom_components/ble_monitor/__init__.py#L105 , subsequent restart will just ignore yaml, because the initial data was put to "options", which is master.
But that can be changed easily.

Re reload, any changes in the UI actually trigger a async_unload_entry and a async_setup_entry in the helpers, with the updated entry

koying:
@Ernst79 So config_flow is the +1 branch and I should PR to that one?

Ernst79:
> So config_flow is the +1 branch and I should PR to that one?

Yes, please. I granted you direct write access to it, if I'm correct. 

I've done some cleaning of imports etc. Please check whether I made mistakes. Not sure if you can modify my first post, if you want things to add to the ToDo list. Otherwise let me know.

Magalex2x14:
@koying I understand correctly that "any changes in the UI" should be understood as clicking a "Submit" button?

koying:
> I understand correctly that "any changes in the UI" should be understood as clicking a "Submit" button?

Indeed :)

Magalex2x14:
Apparently here is a case of multiple entries:

Ernst79:
@koying Just thinking, in the initial integration setup (in the UI), I think we should not have the add_devices flow yet. A user should setup the integration first. After clicking submit, HA will add devices when they are recognized. Next, a user should be able to set the options for these devices from the options menu (options flow). In this menu, we should be able to create a drop down menu with the available devices and set parameters for each device. What do you think?

Ernst79:
Merged the latest changes from @Magalex2x14 and myself from the binary_sensor branch.

koying:
> Next, a user should be able to set the options for these devices from the options menu (options flow)

Won't changing the device name also change the entity name ?

Ernst79:
> Won't changing the device name also change the entity name ?

Yes, at the moment, it will create a new entity with a new entity_id. The old one will be left over (unavailable state) and has to be deleted manually. But that is also the case when changing YAML. After clicking submit, you will return to the option menu with the new entities (and the old one).  I understand this is not ideal, it would be better to delete the old entity automagically, but that is than also needed for YAML, I think.

Magalex2x14:
But we would like to remove the name from the unique ID and rely on the mac address. Or postponed? I don't remember ...

Magalex2x14:
@koying just in case: I suspect that sometimes it is difficult to understand me correctly :) The fact is that my English is a machine translation and intuition sometimes :)

Ernst79:
I did look at only assinging the mac address to the unique_id in the past, but I remember having difficulties with changing the name from YAML than. We ended up with the current situation, as changing the unique_id forced HA to adopt the new name. I remember that I tried it with the unique_id set to the MAC address, but in that case, changing the name in YAML didn't force HA to update the name after a restart. I think because it is stored in the entity registry and can also be overruled manually in the UI. If changing the name from YAML would have worked, it would also imply that manually adjusted names would be overwritten after each restart.

Magalex2x14:
Perhaps this behaviour was due to the fact that before our entities after the restart were created only after the appearance of the first valuable advertisement. Now, this happens much faster with any packet, even if it does not contain a payload. Later, I will also try to experiment...

koying:
Unless the device "name" was always meant to set the entity name, we could set the "friendly" name from it only, and set the entity name from mac,, or type+mac to at least give a hint of what device it is.

That would be a breaking change, ofc.

koying:
To get back to whether to add devices in config flow or not, if discovery is false and we don't have the possibility to add devices in the config flow, we get to a point where we never have devices/entities, if not mistaken.

Ernst79:
> Unless the device "name" was always meant to set the entity name,

It was meant for both the entity name as well as the friendly name. People wanted that their sensors were named e.g. `sensor.ble_temperature_livingroom` for easy recognition.

Ernst79:
> To get back to whether to add devices in config flow or not, if discovery is false and we don't have the possibility to add devices in the config flow, we get to a point where we never have devices/entities, if not mistaken

That's a good point, yes. So let's keep and option to add devices manually in the setup flow, and in the option flow we should have both modifying discovered devices and adding non-discovered devices

Ernst79:
I think you have accidentally deleted `strings.json` in your latest commit.

koying:
> I think you have accidentally deleted `strings.json` in your latest commit.

It seems to (no more?) be necessary. Having only the translation seems to work fine.
Not sure what its use is, really. But yeah, will re-add it.

Ernst79:
According to the developers docs, it is required, but perhaps this has changed? I would leave it in for now, if we aren't sure.

Ernst79:
I've made a more extended abort message. Also added a short description on the forms. 

image

koying:
I consider the config flow done, now.
Having YAML as master makes the code uglier that it should, because HA *really* doesn't make it easy ;)
The device handling is hacky as hell for the same reason.

Now, bug hunting/regression testing/improvement suggestions are welcome :)

Ernst79:
I will start testing. I really appreciate your effort in creating this improvement!

Magalex2x14:
I subscribe to the words of @Ernst79. Thank you!

Before we move on, I want to ask you a question. Can you name any potential advantages and disadvantages of switching to async? In the case of our component, of course.  I managed to run it in the hass event loop yesterday.
So I sit and think why we needed it, and what it is fraught with :)

Ernst79:
Found a bug. When using YAML and setting a device to F, it now shows the temperature in F in home assistant. In the old situation, it was converted to °C automatically, when °C was set as the system unit.

old situation sensors sends 20 --> ble monitor assumes it is 20 F --> HA is set to °C as system temperature unit --> it displays -7°C
new situation sensors sends 20 --> ble monitor assumes it is 20 F --> HA is set to °C as system temperature unit --> it displays 20 F

I think it should be like in the old situation

image

koying:
> Can you name any potential advantages and disadvantages of switching to async

No clue. I'm not quite a Python guru, tbh
The rest of HA seems to go async, so it probably is easier to find examples :)

Ernst79:
In the old situation we had 

`vol.Optional(CONF_TEMPERATURE_UNIT): cv.temperature_unit,`

Which links C or F to °C (= TEMP_CELCIUS) or °F (= TEMP_FAHRENHEIT)

See /homeassistant/helpers/config_validation.py
```
def temperature_unit(value: Any) -> str:
    """Validate and transform temperature unit."""
    value = str(value).upper()
    if value == "C":
        return TEMP_CELSIUS
    if value == "F":
        return TEMP_FAHRENHEIT
    raise vol.Invalid("invalid temperature unit (expected C or F)")
```

koying:
Yeah. Fixed already ;)

Ernst79:
I'll make a list with things I've checked in the start topic.

Ernst79:
We need to think about a logo, I guess. We can just use a Xiaomi of Bluetooth logo, or we can be creative, and make something ourselves.

koying:
> We can just use a Xiaomi of Bluetooth logo

I guess those are copyrighted. Not really sure what is HA position on this...

Ernst79:
They use it all over the place. This is their statement

Trademark Legal Notices
All product names, trademarks and registered trademarks in the images in this repository, are property of their respective owners. All images in this repository are used by the Home Assistant project for identification purposes only.

The use of these names, trademarks and brands appearing in these image files, do not imply endorsement.

Magalex2x14:
There in the repository, there are logos for bluetooth_le_tracker.
And the bluetooth website has a logo pack for download in the media section.

Ernst79:
Another bug/request. When using YAML, I changed a device name, restarted. After that, I deleted the old entities, but it isn't possible to delete the device. Restarting doesn't help, it keeps the old device in the device registry. 

I think we need to use the option flow, when using YAML, to delete devices. Is that possible?

See also this forum [topic](https://community.home-assistant.io/t/delete-individual-device-and-entity-from-integration/157428/40)

koying:
ble_monitor svg-g3732-4294966697

koying:
@Ernst79 the `deconz.remove_orphaned_entries` mentioned in the thread seems a trail to follow.
The tricky part might be to actually identify the "old" entities. If we manage, removing the devices w/o entities seems trivial.

Ernst79:
Yes, looks like the way to go. This is the PR https://github.com/home-assistant/core/pull/40629

koying:
Added a service to delete dangling devices.
Not really sure we can do the same for entities, due to the inherently asynchronicity of the integration

Ernst79:
Service seems to work fine. Entities can be easily deleted manually by selecting the device and clicking on the entities and pressing delete (or from the developer tools, states menu). So I don't think that's a must. 

I've cached another bug, I had YAML config with `decimals: 0`. This works fine. Then I deleted most of my YAML (except `ble_monitor` to return to default, but the number of decimals seems to stay 0, in stead of the default (= 1)

image

Ernst79:
The RFXTRX integration has an option flow that can delete devices. Might be an even better solution it this is possible. But we can also look into this later. 

image

koying:
Indeed, but that won"t work for YAML.

Ernst79:
Ah, that's a pity. I thought we could tell the user if they have orphaned devices after a name change, they could go to the option flow and delete the device there (even in YAML mode), instead of going to the services menu. I find the services menu a bit more "user unfriendly". Isn't it just a matter of deleting devices from the device registry?

I haven't looked at setting up and working with the UI, will do that later, but I expect that we need something there too, to delete devices (as said, I haven't looked at the UI setup, I've started with the YAML setup).

koying:
That would be tricky, as if we allow the user to go to options in YAML, we cannot prevent him to change settings.
If we then overwrite those with the YAML, it would be quite counter-intuitive. as well.

If think the fact that we cannot delete devices is an oversight. I expect HA to add this sooner or latter, and the service is just a temporary solution.

Honestly, the "YAML is master" is already a PITA. Let's not pile up hacks :)

Ernst79:
Yes, I mean just only show the "delete" option in the UI, when using YAML. I understand it's another hack, but is increases the user friendliness. 

About the YAML is master, I understand this complicates things, but we have three choices I guess, 
- UI only (preferred option of Home Assistant)
- YAML and UI. YAML as master. (current option where we are aiming at)
- YAML and UI. UI as master

To my opinion, the 3rd option isn't workable. You set things in YAML, that are overruled in the UI. Changes in YAML wont affect the behavior anymore. Seems not to be a good solution. 

UI only would be the best option, and will be required if we ever decide to move the code to HA. But it will also affect a lot of users who prefer YAML only. 

If I had to choose, I would go for a transition with "YAML and UI, YAML as master" for now. In the future, we can phase out YAML, to get in line with HA requirements. But I have the feeling that @Magalex2x14 isn't in favor of phasing out YAML?

koying:
> To my opinion, the 3rd option isn't workable

1 & 3 are the same really. HA is fine with importing the yaml config, but only once.

> But it will also affect a lot of users who prefer YAML only.

I'm one of those, tbh, and it's quite counter-intuitive that YAML is ignored when an integration moves to config flow.
Problem is that HA makes it quite hard to have YAML as master ;)

koying:
Note that even in a pure UI config, you can still sneakily change the .storage/core_config.entries file to have a yaml-like, config-as-code, configuration :)
Same for devices, btw, in .storage/core.device_registry

If you see it that way, HA just moved the YAML config elsewhere and added UI ob top, which is not so bad.

Ernst79:
I've started testing "UI mode", but I think we need a "delete device" option there as well, we can now only add devices, not delete them. In the ideal situation, it should do two things, if possible, 
- remove it from the config, so the sensor entities become `unavailable`
- remove the orphaned/dangled entities and devices (funny words for a non-English speaking person).

Ernst79:
Sorry to be such a PITA, but another request. Is it possible to show a list of discovered devices in the devices list (in the option flow)? I installed the integration from scratch in the UI, After clicking submit, my sensors got recognized (2 devices are available, with data), but in the option flow, I can't select the discovered devices. 

image

koying:
@Ernst79 Please keep in mind that the config flow is, well, for config :)
View it primarily as a UI on the YAML.

So yeah, adding discovered devices might be a nice addition, and is probably possible via more hacks, but we're twisting the scope of the config flow, here...

Ernst79:
You're absolutely right. I'll add it to the "nice to have" list. 

I wouldn't describe it as a hack, it's just helping the user to submit the config :) My proposal is actually nothing more that asking HA: Look into the device registry, give me a list of the ble_monitor devices and present them as a prefilled list that a user can select when making his config in the UI. so I would call it "helping the user with the config flow". 

But yes, I agree, its a nice to have feature, for "lazy" people who do not have to look up their MAC address. I'll put it on the nice to have list.

Ernst79:
One more, there doesn't seem to be any validation on the YAML config anymore, e.g. I could just load a wrong mac address.

`'devices': [{'mac': '46545464B', 'name': 'slaapkamer3'},`

```
2020-11-30 19:11:18 DEBUG (MainThread) [custom_components.ble_monitor] async_setup_entry: {'rounding': True, 'decimals': 0, 'active_scan': True, 'hci_interface': [0], 'period': 40, 'log_spikes': True, 'use_median': False, 'batt_entities': True, 'discovery': True, 'restore_state': True, 'report_unknown': True, 'devices': [{'mac': '46545464B', 'name': 'slaapkamer3'}, {'mac': 'c4:7c:8d:6b:4f:f3', 'name': 'kelder3', 'temperature_unit': '°F'}], 'is_flow': False}
```

koying:
Yeah. I removed it on wrong assumptions, I think.
I'll put it back.

Ernst79:
Thanks, I think the testing is done.  I'll start working on the documentation and the logo tomorrow.

Magalex2x14:
Maybe let's leave a simple BT logo for now?

Ernst79:
image

The two options for the logo. I think the left option is better, as in the right option, we have 3 times BLE monitor as text.

Magalex2x14:
Yes, I agree. We can then expand the text "BLE Monitor" in the second row to "BLE Monitor and Tracker", for example.
[Link](https://3pl46c46ctx02p7rzdsvsg21-wpengine.netdna-ssl.com/wp-content/uploads/2019/10/Bluetooth-Brand-Guides-and-Logos-English-1.zip) to the BT logo source

Ernst79:
@koying All checkboxes are checked :) Do you also consider it finished for an initial beta release, or do you have more changes coming up. 

@Magalex2x14. As far as I can see, everything works. Do you want to do a final review, perhaps with a focus on the code?

There are a few "nice to have" things (see opening topic), but I propose we keep it as it is now. We can add these things later, in small updates, if we want. I will create a separate issue for that after the merge. In the meantime, you can focus on your device tracker extension. 

I will update the docs and logo tonight.

Magalex2x14:
Yes, I'll take a look later today and try. I need to figure it out, because I suppose there will be an influx of the issues with this release )

koying:
> I think the left option is better, as in the right option, we have 3 times BLE monitor as text.

Agreed, but maybe use a BLE specific logo like https://www.freepng.fr/png-vaoh97/, to differentiate from "plain" bluetooth ?

koying:
> Do you also consider it finished for an initial beta release

Yes. If it's deemed fine by you and @Magalex2x14 , let's get user feedback ;)

Ernst79:
@Magalex2x14 `strings.json` must be the same as `en.json` (the last is the English "translation")

Magalex2x14:
Yes, I noticed that they are the same :) you beat me :)

Magalex2x14:
I tried it. There seems to be nothing to add to what has already been discussed. Let's see how it goes.
It's strange, seems to me, they have it all organized with an UI. I do not like. Perhaps the availability of documentation would close all the questions. Therefore, probably only a few integrations work with the "new" device tracker. The vast majority are still with the legacy...

Ernst79:
I noticed in the last update of Home Assistant that a lot of integrations were removing their YAML support, referring to [ADR-0010](https://github.com/home-assistant/architecture/blob/master/adr/0010-integration-configuration.md#decision). I maintain the official `solarlog` integration in HA (and I am probably the only user), and have also decided to do it. But for ble_monitor, let's just keep the YAML besides the UI. Personally, I like UI configuration. But, that's the advantage of having a "custom component", we can do both.

Ernst79:
> Agreed, but maybe use a BLE specific logo like https://www.freepng.fr/png-vaoh97/, to differentiate from "plain" bluetooth ?

@Magalex2x14 Is this logo also fine by you?

Magalex2x14:
As far as I could find out, there is no specific logo for Bluetooth LE. Everything that is found on the network about this is someone's handicrafts. The only thing that they tried to do for differentiation was the introduction of the "Bluetooth smart", but then they refused. In the archive, the link to which I gave above, there is an interesting file with the rules for the use of logos.

If you like, then put this logo. I would leave the official one, and would think about describing the integration. For example, in the second line of the description ("Entry name"), we can decipher the BLE abbreviation.
image

Ernst79:
I will ask on the github page for the logo's. I had some bad experience in the past with copyright claims of a copied photo, so I'm holding back a bit. A missing logo doesn't have to prevent us from making a beta release. If everybody agrees, I will make a beta release tomorrow.

Magalex2x14:
Bluetooth or Bluetooth le trackers uses official logo. If memory serves me. They are in that repository. In addition, that archive is posted on the Bluetooth SIG site specifically for digital and printed usage. But you better ask, yes. I maybe wrong.

Magalex2x14:
Yes, I think we are ready for beta.

Ernst79:
If I read this, it is pretty clear it isn't allowed

The Bluetooth word mark and **logos can only be used by Bluetooth SIG members**, who are also licensees, and in relation to all qualified Bluetooth enabled goods and services. This includes qualified products, product packaging, sales literature, advertisements, press materials, radio/TV commercials, exhibition stands, outdoor advertisements, **websites**, etc.

Ernst79:
It's released as 0.9.0-beta. 

@Magalex2x14 I've asked about the logo on discord. If I get a response, I will let you know, and we will decide what to do.

Ernst79:
Ok, I've checked it and got a response from Balloob himself. It's ok to upload the Bluetooth logo.

Magalex2x14:
Thank you.

Magalex2x14:
Maybe, since we now in the UI, should we make a notification in HA instead of a warning in the log?
It might also be worth making a notification when a new device is found?

Magalex2x14:
In general, it seems there is some kind of mechanism specifically for notifying about discovered devices. The UPNP integration seems to work like this.

Ernst79:
Yes, I think something like that should be possible. I'll put it on the nice to have list as well. If the device is recognized right after the setup, I also get a screen which shows the device (where I can select the Area). So there is already some kind of trigger.

koying:
I suspect there is a EVENT_DEVICE_REGISTRY_UPDATED autogenerated whenever a new device is created

Ernst79:
PR Logo https://github.com/home-assistant/brands/pull/2063

I also made a PR for this (and I quote)
_PS: Did you know you can also add your custom integration to our Python wheels repository? It will make the installation of your custom integration in Home Assistant lightning fast!_

https://github.com/home-assistant/wheels-custom-integrations/pull/197

Magalex2x14:
@koying, correct me if I'm wrong somewhere.

 - the list of devices displayed in the menu is stored in `config_entry.options.devices` (and flow methods get it from there to display in the UI)
 - the only place with the list of "real" devices actually detected by our integration is the device registry, where entries are created when the corresponding entities appear (binded by identifiers of the `device_info` property)

If everything is so, then in order not to iterate over the entire device registry every time it is updated (not only because of our component), I see a way out in the call to update config_entry from the `added_to_hass` method of the entity. Since this method runs in an asynchronous context, we won't have any problems with sharing the device list. We get the current list, and if the mac-address of the entity is not in the list, then we add it and call the update of the config_entry.

Magalex2x14:
But first, I think it's worth to solve the display of names instead of addresses. I get lost on my list but I don't have many devices...

Magalex2x14:
And think about possibility to display friendly_name, if present. We probably can't do without an event handler here...

Ernst79:
This should show names in stead of mac addresses. Please check if it works for you.

Ernst79:
I now get the icon in the menu where you can install the integration, but not in the integration tile. I thought it should use the icon when there isn't a logo (that's what was explained). I'll wait for a day, otherwise I will upload a logo as well. 

image

image

Magalex2x14:
I was wrong, not a friendly name, but a name by user (in the device registry):
image

Magalex2x14:
In general, why do we need a name field in the UI? Renaming devices is conveniently done - all entities are renamed at once.

Magalex2x14:
> but not in the integration tile

For me logo appeared once, but after the next restart it disappeared again...

Guys, I seriously think that we should remove the ability to change the name from the config flow. This is confusing. Just continue to store the name (for displaying in the UI), but do not let the user change it. Let's rename from the HA device list.

Ernst79:
> In general, why do we need a name field in the UI? Renaming devices is conveniently done - all entities are renamed at once.

Your right, nice catch. You can already rename devices in the UI. We should use that. (It works) an this won’t confuse users. Even better, you won’t get old left over devices and entities!

My earlier change wasn’t right, you can have sensors that don’t have a name set in config, but e.g.only an encryption key. With my latest change, it won’t appear I guess. We just have to append all devices in the ble_monitor domain to the selection list. I’ll look into it tonight.

Magalex2x14:
Okay, I'll focus on supporting new devices then.

Ernst79:
Yes, that's fine. 

Just thinking further, we should/can remove the `name` parameter also from YAML. Not right now, but after a while. 
After a user has updated, sensor entities will get the new device_info, and will show the device in the UI, also in YAML mode. Changing names of sensors can then be done in the same way, in the devices menu in the UI. No need to set it in YAML. All sensors from one device will be changed in one action. No more old, remaining left over sensor entities and devices.  Only drawback is after a reinstall, you have to do this again.

koying:
Does changing the device name in the UI also changes the sensor names? That was the original use case, if not mistaken.
I used the MAC address to display in the dropdown because that's the actual key. Names are fine, ofc, as long as they're unique.

@Magalex2x14 Everything displayed in the config flow actually comes from core.config_entries, currently. That's 1-to-1 with the "devices" in YAML

Ernst79:
> Does changing the device name in the UI also changes the sensor names?

Yes, it does apparently. You will get this message when changing the device name in the UI.

image

Just click rename, and all sensors are changed.

Ernst79:
I think the whitelist option also is not needed anymore. It is a built in feature of HA to disable sensors. Advantage that this is at sensor level, so you can even choose which individual sensor you don't want to show. 

image

I'll look into this as well.

Ernst79:
Name option in UI is removed + fix in temperature_unit in YAML mode

Device list is reverted to MAC address for now, will change that later to device list from registry. 

One "problem", when a user moves from YAML mode to UI mode, it uses the old `name` as `device_name` (OK). But when opening the "device option menu", it will be changed to the default device name (MAC address without : ). As it will remove the `name` tag upon submit.

`{'mac': 'c4:7c:8d:6b:4f:f3', 'temperature_unit': '°F', 'encryption_key': ''}`

Magalex2x14:
I wouldn't be so sure about the whitelist. I suspect that if we remove it, then a scream will rise :) Neighboring sensors, that's all... I would leave the option to disable discovery.

Ernst79:
ok, sure. No problem.

Magalex2x14:
The idea was born - what do you think about ability to export integration configuration in yaml format?

Ernst79:
To export the UI config to YAML? Not sure that is even possible?

Magalex2x14:
It seems to me that it is possible. It is not necessary to write a file, we can just show the text. I'll have to think :)
After new devices, and tracker, and everything above... Nothing, New Year's holidays are coming soon :)

I just decided to reinstall the integration, and forgot that everything disappear :) It's good that my encryption keys are stored on another system.

Ernst79:
> Will look into this later today, it might also happen in UI mode. I guess we need a check to only change the device name in YAML mode or something like that.

Check it, does not happen when in UI mode. So only an issue when moving from YAML to UI. Not sure I understand why this only happens after moving from YAML to UI and not in UI mode. Have to dig further. Perhaps leave it as it is, is OK. 

> New Year's holidays are coming soon
Yep, the list is getting longer and longer. Luckily they are thinking about a third Christmas holiday week here in the Netherlands, due to Covid19.

Ernst79:
@koying Do you know if it is possible to change the code below to an async variant?

```
    @callback
    def _show_user_form(self, step_id=None, schema=None, errors=None):
        option_devices = []
        option_devices.append(OPTION_LIST_DEVICE)
        option_devices.append(OPTION_ADD_DEVICE)
        for device in self._devices:
            option_devices.append(device.get(CONF_MAC))
        config_schema = schema.extend({
            vol.Optional(CONF_DEVICES, default=OPTION_LIST_DEVICE): vol.In(option_devices),
        })
        return self.async_show_form(
            step_id=step_id, data_schema=config_schema, errors=errors or {}
        )
```

(I think) I need this to be async, as I want to get the devices from the device_registry, and this has to happen in an async way, as far as understand. E.g. `device_registry = await async_get_device_registry(self.hass)` and then getting the devices from the ble_monitor domain. 

But changing `def _show_user_form(self, step_id=None, schema=None, errors=None):` to `async def async_show_user_form(self, step_id=None, schema=None, errors=None):` and of course changing the `_show_user_form` to `async_show_user_form` gives me a `TypeError: 'coroutine' object is not subscriptable`. Even removing the option_devices and config schema out gives the same error. Unfortunately, the error doesn't give me clues, except the result["type"] is causing a TypeError. 

```
2020-12-04 11:45:29 ERROR (MainThread) [aiohttp.server] Error handling request
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
resp = await self._request_handler(request)
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_app.py", line 499, in _handle
resp = await handler(request)
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_middlewares.py", line 118, in impl
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 18, in request_context_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 72, in ban_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 127, in auth_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/view.py", line 129, in handle
result = await result
File "/usr/src/homeassistant/homeassistant/components/config/config_entries.py", line 210, in post
return await super().post(request)
File "/usr/src/homeassistant/homeassistant/components/http/data_validator.py", line 60, in wrapper
result = await method(view, request, *args, **kwargs)
File "/usr/src/homeassistant/homeassistant/helpers/data_entry_flow.py", line 69, in post
result = await self._flow_mgr.async_init(
File "/usr/src/homeassistant/homeassistant/data_entry_flow.py", line 130, in async_init
result = await self._async_handle_step(
File "/usr/src/homeassistant/homeassistant/data_entry_flow.py", line 226, in _async_handle_step
if result["type"] not in (
TypeError: 'coroutine' object is not subscriptable
```

koying:
@Ernst79 Could you show your code? We are already out of ble_monitor in your stacktrace.
config_flow is peculiar, when you do `return self.async_show_form`, you're actually returning a dict to the flow manager.

From a config flow, you basically can only return `async_show_form`, `async_abort` or `async_create_entry`

Ernst79:
Yes, I understand. My aim is to fill the option_devices list not with `CONF_MAC`, but with device_names (or `name_by_user`) from the device registry. To do that, I need the query the device_registry, which has to be done async (device_registry = await async_get_device_registry(self.hass)`. Haven't looked at how to actually get the devices from the ble_domain yet, but I will figure that out later myself. But when I add an async function in the code below, it complains the function `_show_user_form` isn't async. So, with my limited knowledge, I thought, lets make `_show_user_form` async. But that didn't work, with the above error. 

Note that I'm not a real programmer, I just try by looking how others do this, and try to figure out by trial and error. But I failed on this one. 

Current code
```
    @callback
    def _show_user_form(self, step_id=None, schema=None, errors=None):
        option_devices = []
        option_devices.append(OPTION_LIST_DEVICE)
        option_devices.append(OPTION_ADD_DEVICE)
        for device in self._devices:
            option_devices.append(device.get(CONF_MAC))
        config_schema = schema.extend({
            vol.Optional(CONF_DEVICES, default=OPTION_LIST_DEVICE): vol.In(option_devices),
        })
        return self.async_show_form(
            step_id=step_id, data_schema=config_schema, errors=errors or {}
        )
```

Async variant (not working)

```
    @callback
    async def aync_show_user_form(self, step_id=None, schema=None, errors=None):
        option_devices = []
        option_devices.append(OPTION_LIST_DEVICE)
        option_devices.append(OPTION_ADD_DEVICE)
        for device in self._devices:
            option_devices.append(device.get(CONF_MAC))
        config_schema = schema.extend({
            vol.Optional(CONF_DEVICES, default=OPTION_LIST_DEVICE): vol.In(option_devices),
        })
        return self.async_show_form(
            step_id=step_id, data_schema=config_schema, errors=errors or {}
        )
```

The non-working code is in the config_flow_ernst branch (only config_flow.py is changed)

https://github.com/custom-components/ble_monitor/blob/config_flow_ernst/custom_components/ble_monitor/config_flow.py

Magalex2x14:
If I understand everything correctly, a callback cannot be a coroutine. I think you can try [this](https://developers.home-assistant.io/docs/asyncio_working_with_async/#calling-async-functions-from-threads).
But I'm not sure if this is the best option. It might be better to keep listing our own device list and just find a way to keep it in sync with the device registry? In the case of the disabled discovery option, changes in the registry are only possible by the user (you can catch it by event, as @koying mentioned above - you just need to write a small event handler, which will compare the contents of the registry with our list, and add devices / update name to our list if required). In the worst case, in such a handler, you will have to iterate through the entire registry, but you need to read about the event (maybe only the changed part of the registry comes to handler), or there is a way to quickly get only our part of the registry.

koying:
You can check the code of the "cleanup" service, which extracts our devices from the registry (then compares it with what we currently have, and removes the diff)

Ernst79:
Sorry guys, I give up. I'm too much of an amateur, I guess. 

I'm only trying to get some async function inside the `_show_user_form` function. Tried the method of @Magalex2x14 , but now it can't find the async function. 

```
  File "/config/custom_components/ble_monitor/config_flow.py", line 132, in _show_user_form
    async_get_device_list(self), hass.loop
NameError: name 'async_get_device_list' is not defined
```

I guess I can better leave this to the pro's. Updated code is here.

https://github.com/custom-components/ble_monitor/blob/config_flow_ernst/custom_components/ble_monitor/config_flow.py

Magalex2x14:
I'll try to look later today

koying:
Not super sure of the logic, tbh.

We have a "configured" list of devices in config_entry, with MAC, and, e.g. encryption key
We have a list of device name in the registry, without MAC but with name.

How do you see the merge of the two in the config_flow?
Getting the device list could be done in https://github.com/custom-components/ble_monitor/blob/config_flow/custom_components/ble_monitor/config_flow.py#L305 , but what do we do with thereafter?

koying:
Ah, but we could return the MAC in the identifiers of the device_info...

Magalex2x14:
The registry contains mac addres. In our case, this is one of the identifiers. Below is one of the registry entries (from `.storage/core.device_registry`):
```json
{
                "config_entries": [
                    "581d6ebeff8684a15afb74aeeb93df0c"
                ],
                "connections": [],
                "identifiers": [
                    [
                        "ble_monitor",
                        "C47C8D6A81BA"
                    ]
                ],
                "manufacturer": "Xiaomi",
                "model": "HHCCJCY01",
                "name": "C47C8D6A81BA",
                "sw_version": null,
                "entry_type": null,
                "id": "1c397a253167abd964c50c13941bcbea",
                "via_device_id": null,
                "area_id": null,
                "name_by_user": null
},
```

Also, I think we can rely on the `name` field, because after importing from yaml, the only way to rename the device is the HA UI, and it does not change the `name`, but the `name_by_user` field.

Magalex2x14:
But identifier is more reliable, yes.

Ernst79:
At this moment, there is an issue that after switching from YAML to config flow, the device is called as defined with YAML (which is OK). But after opening the device option menu of a certain sensor, the device is renamed back to the the mac address field, as the CONF_NAME is removed upon submit. I thing we need to add something that writes the YAML device name to the `name_by_user` attribute and always keep the `name` attribute equal to the MAC address, to prevent this.

koying:
> The registry contains mac addres. In our case, this is one of the identifiers
That's just because the mac address is the default name, currently. But yeah, we can *actually* put the mac address in the identifiers.

Ernst79:
Not entirely true, in YAML mode, it uses the sensor_name, see left picture.

image

koying:
@Ernst79 I think this is what you were trying to achieve

Ernst79:
I will have a look tomorrow, busy tonight.

Ernst79:
@koying Thanks. I had a good night sleep and I'm (finally) getting your and @Magalex2x14 point(s) :-). 
The confusion I had is that the devices drop down list is having two purposes. 

1. Making a list of devices that are used in combination with the `discovery` option
2. Selecting a device to modify the device settings

I was focused on the 2nd, for which you preferably have **all** devices in a list, even those that you haven't added yourself. 
For the 1st purpose, you only want to have the devices that you want to be discovered (when using `discovery: True`). 

Of course, the 1st purpose is the main purpose of this setting. 
So, I have to think this over. For now, I will remove your code (the part that adds devices to the list) and will experiment locally. 

The MAC identifier solves the issue when moving from YAML to UI mode. I will keep that in.

koying:
Actually, I had a good night sleep as well and I think we can have best of both worlds:
- Devices that were configured
+ Devices that had been discovered, by name if modified in HA, thus allowing them to be modified (encryption key, temp unit)

Ernst79:
Yes, I agree. But doesn't it depend on the discovery option?

User that wants to use discovery: False
- List of only devices that he needs

User that wants to use discovery: True
- List of all devices, including the autodiscovered ones. 

How do you see this, one list for both options? or perhaps two options, a multiselect checkbox list (like HCI_interface) for devices to be included in the whitelist, and a add/modify devices list with all devices to modify settings. 

Whatever we choose, in all cases it would be good to use the device name for easy recognition, if available, in the list(s). 
Note that YAML and UI mode are currently dealing slightly different with the name of devices. 

image

With your latest change it isn't duplicated anymore upon a change of the device options, as the identifier now stays the same, which is good. But a device name in YAML is added to the "name" field (left picture), in UI mode (after a new installation of the integration) the name defined in the initial setup is set to the "name" field, later changes of the device name are however added to the "name_by_user" field, the "name" field remains as it was (right picture). It isn't however possible to set the "name_by_user" directly in "device_info" (tried that). Perhaps something with `dr.async_update_device(device.id, name_by_user=custom_name)` .

Magalex2x14:
I didn't sleep very well today, so it's hard to follow your thoughts :)
But here are some of my thoughts:

 - we can leave the imported from yaml name instead of the mac-address. It is unique within our domain.

 - Since tuples are used in identifiers, and tuples are inmutable, they can be used as an index in a list and compared. You can try this code online (https://www.tutorialspoint.com/execute_python3_online.php):
```python
DOMAIN = "ble_monitor"
name_or_mac = "name_or_mac"

device_identifiers = [("other_domain", "other_device"), ("ble_monitor", "name_or_mac")]

#  for mac in our_devices:
ourid_tuple = (DOMAIN, name_or_mac)
ourid_list  = [DOMAIN, name_or_mac]  # even though it looks like this in the registry file, I think the HA core imports it at startup as tuple

print(ourid_tuple in device_identifiers)     # True
print(device_identifiers.index(ourid_tuple)) # 1
print()
print(ourid_list in device_identifiers)      # False
print(device_identifiers.index(ourid_list))  # ValueError: ['ble_monitor', 'name_or_mac'] is not in list
```
----
 - the only source of the device list for the UI rendering procedure in our menu should be our own list (devices). We just need to use an event handler for the EVENT_DEVICE_REGISTRY_UPDATED event to synchronize our list with the device registry changes. Here's what comes in the event:
```
2020-12-04 21:55:36 INFO (MainThread) [custom_components.ble_monitor] event: <Event device_registry_updated[L]: action=update, device_id=23d63478886605994f3a78750db3d7d5>
2020-12-04 21:57:17 INFO (MainThread) [custom_components.ble_monitor] event: <Event device_registry_updated[L]: action=create, device_id=674c969bfd2cb06591a54a36493e8ccc>
```
----
 - our list already contains devices manually added by the user (regardless of the "discovery" option) - there are only mac-addresses (and names if imported from yaml). The event handler will add names_by_user to it (when renaming with UI) and new devices (discovered), in which we just need to check if it is our device has been added/updated, and if ours, then update the list of devices in config_entry with added/updated device data (name_by_user or device itself). Our config_entry:
```
{
                "entry_id": "581d6ebeff8684a15afb74aeeb93df0c",
                "version": 1,
                "domain": "ble_monitor",
                "title": "Bluetooth Low Energy Monitor and device tracker",
                "data": {},
                "options": {
                    "hci_interface": [
                        0,
                        1
                    ],
                    "discovery": true,
                    "active_scan": false,
                    "report_unknown": true,
                    "batt_entities": true,
                    "rounding": false,
                    "decimals": 1,
                    "period": 60,
                    "log_spikes": false,
                    "use_median": false,
                    "restore_state": false,
                    "devices": [],   # <----- I have it empty here, because I did not add devices manually, and did not import from yaml
                    "is_flow": true
                },
                "system_options": {
                    "disable_new_entities": false
                },
                "source": "user",
                "connection_class": "local_push",
                "unique_id": "Bluetooth Low Energy Monitor and device tracker"
            }
```
Since the names in our list are needed only for display, we can safely replace name field with the contents of name_by_user from the registry. And leave the procedure for obtaining a list of devices in show_user_form as is (name instead of mac-address).


P.S. Yes, listening to all registry update events looks a bit expensive, since we only need names when the list of devices is opened in our menu. But from the event handler, we can make notifications about the discovery of a new device (when the device added to the registry but it is not in our list).

Magalex2x14:
We can quickly check if an event is an event about our device by checking if our config_entry_id is in the config_entries list of the updated / added device.
If it is, then we are already going through our list of devices, and checking the addresses / names.
It remains only to figure out how to pass our config_entry object to the handler

Magalex2x14:
I missed solution from @koying. https://github.com/custom-components/ble_monitor/pull/173/commits/d2acc1bf6f5355417661f1d72aa14ca66fed8d1a
I think this is a good solution if it works. At the `async_step_init` point, we have access to `config_entry` and the device registry. And it's not nearly as wasteful compared to an event handler. I just would leave `get_sensorname` in the id. And perhaps I would rework the `for k, v in dev.identifiers` loop, taking into account the fact that we may already have manually added devices in the list that have __not yet been detected__.

@ Ernst79 it doesn't depend on `discovery` option:
 - discovery disabled: only devices that have been added to it manually, plus changes from the registry (name), will be included in our device list. If discovery is disabled, devices other than those added manually cannot get into the registry. Since we only need mac addresses for the scanner to work, we do not need to restart the scanner to update the whitelist, it has all the necessary data.
- discovery enabled: our device list will include devices added manually (with a possible name update from registry), plus everything that was discovered (and possibly renamed with UI). We do not need to restart the scanner because the whitelist is disabled.

The only subtle point is when is `config_entry` saved to file? Don't need to additionally inform the HA core about its update? Simply, let's say, in the event of a power failure or some kind of error, at the next launch of the integration, the platforms can start creating entities with the old names.

Magalex2x14:
I think after processing the list it is worth calling this (since it calls `self._async_schedule_save()` at the end):
http://dev-docs.home-assistant.io/en/master/api/config_entries.html#homeassistant.config_entries.ConfigEntries.async_update_entry
image

koying:
I think I'm there.
Attention point is that if you import a name from YAML, then switch to UI, then edit a named device in UI, you loose the name and have to recreate it from UI-device

koying:
> The only subtle point is when is `config_entry` saved to file

We have a async_update_entry in async_setup_entry, and async_setup_entry is called each time we update via the UI

Ernst79:
> Attention point is that if you import a name from YAML, then switch to UI, then edit a named device in UI, you loose the name and have to recreate it from UI-device

@koying Yes, that was exactly the point I tried to make earlier, YAML uses the `name` field of the device registry to store the custom device name, UI uses the `name_by_user` field of the device registry. After switching from YAML to UI, and subsequently editing a device in the option flow that was named in YAML, it overwrites the name field with the MAC address (without :) as it doesn't have a CONF_NAME anymore. So the user has to do the name change once more in the Home Assistant device menu. 

So, either we accept this, or it might be possible to let YAML use the same principle to define the device_info. (`name = self._mac` and `name_by_user` = self.get_sensorname()`). But the last (setting `name_by_user`) isn't possible directly in `def device_info(self)`. 

Anyway, I will test your changes today, thanks.

koying:
Actually, the name is saved from the YAML config up to the point where we edit the device in the UI.
As the name option is gone from the UI, only then is the name removed from the config_entry, and thus reverted.

We can workaround this by saving the name before going into the UI, then restore it aftewards. We can also detect via the "name_by_user" if the name was overwritten, then ignore the legacy yaml name.
Not a big deal, but more "hacky" code ;)

Ernst79:
Yes, exactly. In my mind, I thought, lets make YAML define the device attributes in the same way as the UI does. But I'm not a python exert at all, I'll let you decide. :-)

Ernst79:
I've found one issue, probably related to the above. I have two sensors configured in YAML. 
```
devices:
  - mac: '4C:65:A8:DD:B8:9B'
    name: Bedroom
    temperature_unit: C
 - mac: 'c4:7c:8d:6b:4f:f3'
   name: Livingroom
   temperature_unit: F
```

Then moved to UI mode. One of the sensors (Livingroom) is displayed twice in the drop down menu. This is the one which had a **lowercase** MAC address in the YAML. 

image

In the device_registry, there are only two devices. 

```
            {
                "config_entries": [
                    "62fe412ed544c736f2dc2c8fcc6890eb"
                ],
                "connections": [],
                "identifiers": [
                    [
                        "ble_monitor",
                        "C4:7C:8D:6B:4F:F3"
                    ]
                ],
                "manufacturer": "Xiaomi",
                "model": "HHCCJCY01",
                "name": "Livingroom",
                "sw_version": null,
                "entry_type": null,
                "id": "42586d35749310885f25878ba167e811",
                "via_device_id": null,
                "area_id": null,
                "name_by_user": null
            },
            {
                "config_entries": [
                    "62fe412ed544c736f2dc2c8fcc6890eb"
                ],
                "connections": [],
                "identifiers": [
                    [
                        "ble_monitor",
                        "4C:65:A8:DD:B8:9B"
                    ]
                ],
                "manufacturer": "Xiaomi",
                "model": "LYWSDCGQ",
                "name": "Bedroom",
                "sw_version": null,
                "entry_type": null,
                "id": "0fe18ad5e6a30711e0f1cd45efd2631f",
                "via_device_id": null,
                "area_id": null,
                "name_by_user": null
            }
```

Ernst79:
@Magalex2x14 Sorry, I did it again. I will revert

Ernst79:
Sorry guys, I'll create a new PR.

PR #184, opened 2020-12-05
Kettle ext_state explained

piotr-kubiak:


Magalex2x14:
Thank you!
To keep you informed - we also have the `info.md` ​​file containing the same text, but intended for displaying information in HACS. There you need to duplicate this change (I did it myself yesterday).

piotr-kubiak:
Sorry, missed that one :)

PR #188, opened 2020-12-06
Config flow

Ernst79:
**Description**
This PR will add a config flow to BLE monitor.

This PR is a merge of the work of @koying (thanks!) (#168) and the `binary_sensor` branch, which is the basis of beta 0.8.4. 

**ToDo list**

- [x] Fix import of devices (and device parameters) from YAML
- [x] Decide on whether YAML import or config flow is most important
  - [x] Disable option flow when YAML is used
  - [x] Disable removing and adding integration when YAML is used (if possible)
- [x] Option flow for changing devices
- [x] Check on only allowing one integration in the UI
- [x] add logo (see instructions [here](https://developers.home-assistant.io/blog/2020/05/08/logos-custom-integrations/))
- [x] Add/modify documentation
- [x] "Manually" validate the MAC address and encryption key format

Feel free to add things to the ToDo list

**Tests**
The following tests have been checked by @Ernst79 
YAML mode
- [x] Import of YAML with all options
- [x] Import of YAML with no/partial options
- [x] Change in YAML changes BLE monitor
- [x] Check what happens on a name change in YAML
         _New device is created, old one is still a device, but sensors are `unavailable`. They can be deleted. **Device can be deleted with service `ble_monitor.cleanup_entries`** Instructions to be added in the docs_
- [x] Check on validation of input (mac, encryption key)

Switching from YAML to UI mode
- [x] Check what happens when commenting out YAML
        _All devices and entities still exist, integration parameters are according to setting as per old YAML, settings are available to change in the UI. Excellent!_

UI mode
- [x] Create integration (user input flow)
- [x] Options (options flow)
- [x] Adding devices (add_device flow)
- [x] Modify integration settings
- [x] Modify device settings
- [x] Check on defaults
- [x] Check on validation of input (mac, encryption key)
- [x] Check on delete and adding an integration while using YAML

Switching from UI mode to YAML
- [x] Check what happens when commenting out YAML
        _It loads the YAML config, as expected_

**Ideas for future enhancements **
Not required before a release, but these ideas can be implemented in the future. 
- [x] Show the device name instead of the MAC address in the `devices` drop down menu
- [x] Automagically prefill the `devices` drop down list with discovered devices for easy configuration
- [ ] Add a `delete device` option in the same drop down menu where you can select `Add device`. It should show a drop down form where you can select devices to delete.
- [ ] We could look into the possibility that clicking submit will delete the orphaned/dangled entities and devices.
- [ ] After deleting a device, it isn't possible to remove the sensor entities. A restart is required before this is possible. Unload entities from the registry would probably solve this. 
- [ ] Add notification when a device is detected


Ernst79:
So, we are back. Sorry about that. 

Difference in number of files in that are changed in this PR (11) and in the previous (15) is the gitignore file and three pictures (which were equal). 

I included the latest changes from @Magalex2x14 in master. 

The files are in the **config_flow_restored** branch now. Please don't use config_flow anymore (will delete them tomorrow, if you confirm that it's safe :). 

Sorry for the inconvenience.

koying:
Anything against the gitignore?
I link my git workdir with HA, so a __pycache__ is created in custom_component for me

Ernst79:
No, of course not. It's back.

Magalex2x14:
If you think it is ready for beta, then I will only merge the code from the kettle branch (with the rest of the sensors, not everything is clear yet).

Ernst79:
Yes, I think its ready. It works fine on my end, lists are with device names if available, so looks good to me. Other improvements can be done later, in small updates. So, please merge the kettle branch in here.

Magalex2x14:
Done.

Magalex2x14:
I have failed with an upgrade on my main system to 0.9.2-beta.
An error is recorded in the log:
```
2020-12-07 03:09:24 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/data_entry_flow.py", line 130, in async_init
result = await self._async_handle_step(
File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/data_entry_flow.py", line 248, in _async_handle_step
result = await self.async_finish_flow(flow, dict(result))
File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/config_entries.py", line 498, in async_finish_flow
entry = ConfigEntry(
File "/srv/homeassistant3.8/lib/python3.8/site-packages/homeassistant/config_entries.py", line 159, in __init__
self.data = MappingProxyType(data)
TypeError: mappingproxy() argument must be a mapping, not builtin_function_or_method
```
The error occurs immediately upon restarting HA.
Trying to add integration to HA throws an error"Unknown error":

My config in yaml:
```yaml
ble_monitor:
  rounding: False
  decimals: 2
  period: 60
  log_spikes: True
  use_median: False
  batt_entities: True
  active_scan: False
  discovery: True
  hci_interface:
    - 0
    - 1
    - 2
  report_unknown: True
  restore_state: False
  devices:
    - mac: '58:2d:34:31:40:00'
      name: 'bed_sensor'
    - mac: '58:2d:34:31:0b:21'
      name: '2'
    - mac: '58:2D:34:31:0A:CB'
      name: '1'
    - mac: 'C4:7C:8D:6A:81:BA'
      name: 'flower'
    - mac: "A4:C1:38:2F:87:6B"
      name: 'external'
      encryption_key: "4F752A547B902714CF312C596F1E7C72"
    - mac: "A4:C1:38:D1:6A:8D"
      encryption_key: "C99D2313182473B38A01086FEBF781BD"
      name: 'office'
    - mac: "E4:AA:EC:34:B6:57"
      encryption_key: "B86B49D5B14613A57B5C5919B2CB417B"
```
Rollback to 0.8.5 solves the problem. I haven’t figured it out yet.
Everything is fine on my test raspberry, it differs in that 0.9.0-beta was first installed on it and the configuration was removed from yaml.

Magalex2x14:
After updating on a test raspberry, I found a mess in the devices - I thought it was because of my tests, and removed the integration. Rebooted HA (entities and devices have cleared). I could not add the integration back - the same error as above.

Magalex2x14:
Reinstalling 0.9.1-beta goes smoothly.

Magalex2x14:
i decided to remove 0.9.2-beta tag (release is drafted)

Magalex2x14:
Upgrading from 0.9.1 to 0.9.2 goes smoothly, but all devices are duplicated. Attempting to reinstall 0.9.2 fails with an unknown error as above, but the log entry on raspberry is different:
```
2020-12-07 03:51:24 ERROR (MainThread) [aiohttp.server] Error handling request
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
resp = await self._request_handler(request)
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_app.py", line 499, in _handle
resp = await handler(request)
File "/usr/local/lib/python3.8/site-packages/aiohttp/web_middlewares.py", line 118, in impl
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/request_context.py", line 18, in request_context_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/ban.py", line 72, in ban_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/auth.py", line 127, in auth_middleware
return await handler(request)
File "/usr/src/homeassistant/homeassistant/components/http/view.py", line 129, in handle
result = await result
File "/usr/src/homeassistant/homeassistant/components/config/config_entries.py", line 169, in post
return await super().post(request, flow_id)
File "/usr/src/homeassistant/homeassistant/components/http/data_validator.py", line 60, in wrapper
result = await method(view, request, *args, **kwargs)
File "/usr/src/homeassistant/homeassistant/helpers/data_entry_flow.py", line 106, in post
result = await self._flow_mgr.async_configure(flow_id, data)
File "/usr/src/homeassistant/homeassistant/data_entry_flow.py", line 155, in async_configure
result = await self._async_handle_step(flow, cur_step["step_id"], user_input)
File "/usr/src/homeassistant/homeassistant/data_entry_flow.py", line 248, in _async_handle_step
result = await self.async_finish_flow(flow, dict(result))
File "/usr/src/homeassistant/homeassistant/config_entries.py", line 498, in async_finish_flow
entry = ConfigEntry(
File "/usr/src/homeassistant/homeassistant/config_entries.py", line 159, in __init__
self.data = MappingProxyType(data)
TypeError: mappingproxy() argument must be a mapping, not builtin_function_or_method
```

Ernst79:
I can confirm this, have the same error when opening the option flow window. the update went smoothly yesterday, but I guess that is related to the fact that I had my modified code already on my Pi. I think the error has is caused somewhere with the merge of the kettle branch, I guess, as I've tested option window before this merge. I guess something doesn't play nicely together.

Ernst79:
Found it, you missed one letter :). Can you check it and rerelease?

Magalex2x14:
Thanks. Felt like this was my commit, but didn't see it in a rush. I'm sorry.
I tested it, everything is fine. Doubled devices - an expected nuisance due to a change in identifiers, resolved with a service. I'll return the release now.

Magalex2x14:
Integration duplicates entities in some situations. Didn't notice when it exactly happened.

I suspect that the problem is related to the fact that we change config entry when generating the list of devices for the menu, which entails changing the result of `get_sensor_name` for an entity that has not yet been added, but already added entities continue to return the old name and unique_id, since `get_sensor_name` is called only when they initialized. The same problem is relevant when renaming in the UI.

The unique ID should not be changed under any circumstances.
When working with yaml, this is true. When working with flow, no.
I think the solution is that when working with flow, there should always be a mac-address in `self._unique_id`.

Magalex2x14:
The problem remains at the moment of switching from yaml-configuration to UI (removal of yaml).

Magalex2x14:
It seems that I managed to solve it. Need deeper testing transition from yaml.
I decided that the easiest way out would be to fix the name as a unique ID when switching from yaml. Thus, after the transition, it becomes possible to rename the imported device with UI. You can only reset a unique ID on a mac by deleting the device.

Ernst79:
Great, I’ll test it, but will do that tomorrow.

Magalex2x14:
Yes, I have a clean transition from yaml. No duplicates or other pain in future.
Unfortunately, this does not apply to those who migrated from yaml with current beta... But this is the overhead of the betatest.

Ernst79:
I tried, but the device is renamed when opening the device configuration window. It's renamed back to the MAC address after clicking submit, sensor entities are duplicated. (it was livingroom in YAML). 

image

koying:
@Ernst79 Would you mind detailing the steps to reproduce, please

Magalex2x14:
@Ernst79 are you sure you made the switch from yaml on fresh code from the branch? As I said above, smooth transition does not apply to those who __migrated__ from yaml with current beta.
You can look at the contents of `.storage/core.config_entries` - there should be unique IDs for all devices that had a name at the time of the transition:
image

Ernst79:
I copied all files from the `config_flow_reversed` branch to HA. 

1. Deleted the integration from the UI
2. Added the YAML code as shown below
3. Restart HA 
4. Checked the devices (livingroom and bedroom, OK)

image

5. Removed YAML code
6. Restart HA
7. Checked the devices (livingroom and bedroom, OK), unique ID is added in registry

image

8. opened the integration settings, opened the device settings of livingroom, changed Fahrenheit to Celsius, submit
9. Checked the devices (C47C8D6B4FF3 and bedroom, NOT OK), **unique ID and name disappeared from registry.** 

image

Perhaps related to lower case mac?

```
ble_monitor:
  rounding: True
  decimals: 0
  active_scan: True
  hci_interface:
    - 0
  period: 40
  log_spikes: True
  use_median: False
  batt_entities: True
  discovery: True
  restore_state: True
  report_unknown: True
  devices:
    - mac: '4C:65:A8:DD:B8:9B'
      name: Bedroom
      temperature_unit: C
    - mac: 'c4:7c:8d:6b:4f:f3'
      name: Livingroom
      temperature_unit: F
```

koying:
Thanks. unique_id is not part of the Device schema as far as config_flow is concerned, so is removed (same as name).

Now, we might be overcomplicating stuff. In sensors, we can just get the device name by MAC.
The device name is created the 1st time the YAML is imported. So it's just there.

Basically same as in config_flow:
```python
        devreg = await self.hass.helpers.device_registry.async_get_registry()
        for dev in device_registry.async_entries_for_config_entry(devreg, self.config_entry.entry_id):
            for iddomain, idmac in dev.identifiers:
                if iddomain != DOMAIN:
                    continue
                name = dev.name_by_user if dev.name_by_user else dev.name
```

Magalex2x14:
@Ernst79 thanks, I was able to reproduce the problem.

I forgot about the schemes yesterday, yes. Thanks for pointing this out. I don't like this New Year's Eve, which usually, in addition to the anticipation of the holiday, brings with it this not very pleasant fuss and mess in business, which confuses thoughts and interferes with concentration :) For some time I worked as a system administrator, and we had a "written in blood" golden rule - not to roll out updates on Friday :) There is something similar, in my opinion :)

There is something in @koying's words, of course. We took the name from the registry, okay. What will we give out as a unique ID? And I didn't really understand what you means about the corner case (translation difficulties, probably), but the problem is not in the device name. The problem is that until now the unique id of the entity was generated based on the name if it was setted in the yaml (which is not correct, and we talked about this earlier, but decided to live with it, since the change was published). I am trying to change this with the transition to the UI, leaving the old behavior for those who want to continue using yaml. It looks difficult at first glance, but the solution is actually simple - to fixate the name if it was set as a unique ID when switching from yaml. All new devices will have a mac address in the entity ID after transition. And in both cases (for old "named" devices and for new ones), there is freedom to change their name through the UI, duplicates will not be generated, since unique entity IDs are fixed.

Magalex2x14:
It looks like it was a matter of overwriting the dev options, not the absence of fields in the schemas.

Magalex2x14:
Try to chase with the last commit as time will be. I don't see any problems yet.

Ernst79:
I checked it again, works fine now. No more duplicates. Excellent! 

I did found another thing, the restore state option seems to do strange things with temperature, when using Fahrenheit and displaying in Celsius (just another edge case). I got peaks of -30°C, in the restore state period. After one period, it return back to normal (measurement of 18F = -9°C. See all the peaks in the picture below. 18°C and -9° are me, switch C to F, but -30°C, I don't get these. 

image

Magalex2x14:
Perhaps it has something to do with None in the restore state? It was with binary entities (I fixed it recently) - None was interpreted as False.

Magalex2x14:
Hmmm. No, there just `self._state = old_state.state`

Ernst79:
It shows -30°C, but I don't get how it can come to such a value. Even converting twice should give -22°C. -22F is = -30°C, so probably applied 3 times.

I tried it again in UI mode, now it shows -22°C, which is applying the conversion 2x. 

`old_state.state` is -9, and it will be converted probably again to Celsius somewhere.

Magalex2x14:
HA core, I guess. And probably we need to do the reverse conversion when restoring state.

Magalex2x14:
And -30 turned out when you restored an entity that was not yet updated before restarting.

Magalex2x14:
Or maybe the point is that we do not restore the unit of measurement?

Ernst79:
One of them is fixed, I now get -22°C during restore state (not -30°C anymore). I get it when I open the device configuration window and click submit. During the restore state period, its still -22°C.

Magalex2x14:
I still don't understand what's going on at all. I cannot get the entity to report in the attributes that it is Fahrenheit. After the first update, the unit of measurement becomes Celsius. What I just didn’t do ...

Ernst79:
No, what Home Assistant does is the following, It receives e.g. 18°F from our component (18 and unit of meausurement = °F). The system unit in Home Assistant is set to °C in configuration --> General --> Unit System. Home Assistant converts the 18°F to -9°C to display everything in °C, even if the sensor sends °F. This is correct.

Magalex2x14:
It looks like in HA everything depends on the global setting.
The core changes the unit of measurement for known units, and all attempts to change this for the entity are ignored. The very first update of the entity sets the unit of measurement in accordance with the global setting. Accordingly, when the state is restored, recalculation occurs if the unit of measurement does not match the system one. This can be checked by setting Fahrenheit globally - then additional recalculation will be done for all sensors except for the one with Fahrenheit in the settings :)

Someone here thought badly, in my opinion :)

Magalex2x14:
@Ernst79 , yes you beat me) As I understand it, there is no error? The device setting must match the global one - then there will be no recalculation when restoring.

Ernst79:
No, If system setting in HA is the same as the sensor setting in ble_monitor, there will be no issue (and this will be the case in 99,9999% probably). When these are different, there is an issue during the restore_state for some reason. When measurements start again after the first period, all is ok. So not a big deal, just don't get it :-)

Magalex2x14:
Yes, I see, you're right :)
It turns out that in order to solve this problem, we need to know in which unit of measurement the last state of the entity was saved. It is necessary to store this setting in the very essence, and in a place that is available during recovery. Then we can compare it with the current global one and recalculate back, if necessary :)

Or is it better, probably, to store the readings in another variable other than `_state` :)
We can restore the state from `mean` or `median`, depending on the presence of the `last mean of` or `last median of` in the attributes :)

Ernst79:
Note that mean or median do not have a Unit of Measurement, so Home Assistant will not do any conversion. So the mean and median can be quite different compared resulting temperature.

Magalex2x14:
No, HA will do conversion. This is the point.

Magalex2x14:
Try last commit :)

Magalex2x14:
When calculating the entity state, the value is taken from the mean or median (depending on the `use_median` setting), so it is safe to recover from them.

Ernst79:
Yes, that helped. It's not staying constant -7.8°C during recover_state, as it should. 

What I meant was the following. Temperature is 18F, displayed as -7,8°C in HA due to my settings (That is OK). But the mean and median are shown in Fahrenheit. 

image

But that was not the issue, it is solved now.

Magalex2x14:
Oh yes, they are displayed in the units reported by the sensor.

Magalex2x14:
@Ernst79 what do you think about returning support for HA versions up to 0.115 in the next release?

Ernst79:
What do you mean? Do we not support HA < 0.115?

Magalex2x14:
I think yes. But come on tomorrow. I have a few more merges to do there, now I’m not thinking well, I’m afraid to forget something.

Yes, we lost support of all until 0.115. There was just a chat on the [forum](https://community.home-assistant.io/t/passive-ble-monitor-integration-xiaomi-mijia-ble-mibeacon-monitor/177352/681?u=magalex). I made a separate branch 0.8.5_pre0.115. It works at 0.114. I haven't tested with current HA version yet, but I think everything will be fine.

Ernst79:
Ok, sure. No hurry. 

And fixing it for old HA versions, I can’t have anything against that. :)

Magalex2x14:
We must remember to write in the release note that for those who switched from yaml (deleted ble_monitor from config.yaml) on previous versions of the 0.9.x branch, entities can be duplicated.

Ernst79:
Aha, the percentage change. Ok, to revert that for now, but I would not recommend to keep support for old versions forever. At a certain moment, we should just advise to update HA.

Magalex2x14:
Yes, and BinarySensorDevice. That's why I asked you what you think about this ) Maybe we really shouldn't do it, but I see no reason for the opposite.
There's a simple `try..except` on import.

Magalex2x14:
@Ernst79 seems to have nothing forgotten.
I added more changes to info.md ​​(0.9.3-beta changes) - please, check the text.

Magalex2x14:
I tested the changes in the code. I don't see any problems.

Ernst79:
I’ll do a last check,

Ernst79:
Yes, I don't see any issues either. YAML and UI both work. Can you do the release? I have to go now.

Magalex2x14:
Yes, shure

Magalex2x14:
Guys, how many users do you think we have? Is it possible to estimate their number? At least approximately...

Ernst79:
I ave no idea, but we have 2500 Unique visitors according to the stats.

https://github.com/custom-components/ble_monitor/graphs/traffic

Magalex2x14:
Yes, I saw this statistics. But only now I noticed that they read our FAQ )
Although ... This is probably for instructions on obtaining encryption keys )

Ernst79:
@Magalex2x14 What do you think, is it ready to move to a final release?

Magalex2x14:
I think yes, it's done. The only thing that bothers me a little is the WildeRNS issue at the end of #75, although I doubt it's a component problem.

Ernst79:
Let's wait then, and see if he response tonight/tomorrow.

Ernst79:
@Magalex2x14 I've looked into a way to get the number of downloads. I found an example in the HACS integration itself. It is showing the number of downloads of its own integration in a badge. They make this badge with shields.io, more specific https://shields.io/category/downloads

So I tried that with the latest release, but it showed 0. 

https://img.shields.io/github/downloads/custom-components/ble_monitor/latest/total
[Downloads for latest release](https://github.com/custom-components/ble_monitor/releases/latest)

So, looked further and checked the API from github, which should provide the data in a `download_count` parameter in `assets`. 
https://api.github.com/repos/custom-components/ble_monitor/releases

In our case `assets` is empty unfortunately. 

```
"assets": [

    ],
```

Turns out that we have to include an additional zip file when we make a release, to make this work. 

https://github.com/badges/shields/issues/833#issuecomment-502971734

So, in our next release, I think we can try this by downloading the repository as a (additional) zip file and add it to the assets of the release. If that works, it should be counting downloads, I think. Only thing I'm doubting about is that the number of HACS downloads seem to be too small (1200). But lets just try it. 

Regarding the next release, I assume the issue with the kettle is related to bad reception. I propose we make the current beta final. I already started to work on bringing a first parameter (`decimals`) to the device level, but I would like to have the current status released first and start to work from there.

Magalex2x14:
Yes, I agree that the problem is not related to the integration work, and we can make a release. I'm a little busy at the moment. Will you do it?

Ernst79:
yes, no problem.

Ernst79:
It now has a `"download_count": 0` parameter in the API, added the badge in the release for testing, lets see if the download badge will work.

Magalex2x14:
@Ernst79 I made a commit to the master with a bug fix when adding devices manually and moved the 0.9.3 tag to this commit. Anyone who encounters this problem will need to reinstall the component in HACS.

PR #199, opened 2020-12-16
Decimals setting at device level

Ernst79:
This PR adds a `decimals` setting at device level, that can overrule the integration setting for individual sensors. Default value is `default`, which means that it will use the integration setting. 

image

Other changes:
- `sensorname` is renamed to `device_name` for example `get_sensorname` is renamed to `get_device_name` as this fits better with the naming of HA. 
- Some commented out code is removed
- Applied `black` on parts of the code to clean it up a little bit. 

Note that the PR is targeting the `decimals-at-device-level` branch, in case you want to modify things. I assume if you merge it, it will get into this branch first, where you can do modifications, if needed, before we merge it into `master`.

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -26,8 +26,9 @@ from homeassistant.helpers.entity_registry import (
 )
 
 # It was decided to temporarily include this file in the integration bundle
-# until the issue with checking the adapter's capabilities is resolved in the official aioblescan repo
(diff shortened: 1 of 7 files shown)


Magalex2x14:
I don't have a chance to look at it yet... If there are any more changes, you can include this code in the next beta, if you are sure of it?

Ernst79:
There is no hurry. Even if you need weeks before you have time, it can wait. I’ll move it to a branch here, and perhaps add more settings to device level. Don’t feel any obligation to look at it soon, there is no urgent need for this.

PR #209, opened 2020-12-20
Update README.md

jcisio:
Fix battery name.


Ernst79:
Thanks @jcisio

jcisio:
No problem. It's the least I can do. Thank you for spending your times making and maintaining the integration, which supports many more devices than what you have, I'm sure!

PR #213, opened 2020-12-22
[WIP] Adding decimals setting at device level

Ernst79:
This PR adds a decimals setting at device level, that can overrule the integration setting for individual sensors. Default value is default, which means that it will use the integration setting.

image

Other changes:

- sensorname is renamed to device_name for example get_sensorname is renamed to get_device_name as this fits better with the naming of HA.
- Some commented out code is removed
- Applied black on parts of the code to clean it up a little bit.

Note. I moved the repository to here (`decimals-at-device-level` branch). I made in [WIP], will try to add/change the following. 

- [ ] Remove `rounding` option

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -26,8 +26,9 @@ from homeassistant.helpers.entity_registry import (
 )
 
 # It was decided to temporarily include this file in the integration bundle
-# until the issue with checking the adapter's capabilities is resolved in the official aioblescan repo
-# see https://github.com/frawau/aioblescan/pull/30, thanks to @vicamo
+# until the issue with checking the adapter's capabilities is resolved in
+# the official aioblescan repo see https://github.com/frawau/aioblescan/pull/30
+# thanks to @vicamo
 from . import aioblescan_ext as aiobs
 
 from .const import (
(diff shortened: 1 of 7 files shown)


Ernst79:
Removal of the rounding option is added. 

It will show a warning, when still in YAML configuration. 

image

Ernst79:
Added the `restore_state` parameter at device level. 

Also added a new `get_device_settings()` function. 
In the old situation, we had two functions, one to determine the temperature unit and one to determine the name of the sensor. In the new situation, for each parameter that can be overruled at device level (`name`, `temperature_unit`) or can be set at device level as well as integration level (`decimals`, `restore_state`), I needed a more or less similar function, so I decided to group them in a `get_device_settings()` function. This function basically sets the parameters to the setting that is defined at integration level, and overwrites it when set different at the device level. The output is a dictionary with all settings for that specific device.

Ernst79:
Added the `use_median` parameter at device level. 

I think this is finished now. 
Other settings like `active_scan` and `log_spikes` are not really relevant at device level. `batt_entities` can be relevant, but I think it is better to remove it, and always show the battery sensor. HA has the built in possibility to disable sensors, if you don't want certain sensors. See also my post in the discussions.

Ernst79:
Do not merge this PR. It is already partly implemented (the structure). Will create a new PR coming week.

PR #230, opened 2021-01-09
Add note for using older MiHome mod as latest does not write pairings…

jeremysherriff:
….txt

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -227,6 +227,7 @@ The BLE messages from some sensors are encrypted. To decrypt these messages, you
 
     - You will first have to allow apps to be installed from “unknown sources” in your android settings (Settings – security – select “unknow sources).
     - [Download](https://www.kapiba.ru/2017/11/mi-home.html) MiHome mod and install the MiHome mod apk on your android phone. Use google translate to translate the website, as the website is only available in Russian. At the bottom, you will find the download link to the latest version. MiHome mod is basically just MiHome with some translatons from chinese to English and Russian, but also with the possibility to save the encryption key.
+    - You may need to use an older version - visit the "ARCHIVE previos versions" area. Tested as working on MiHome_5.7.5_63055_vevs.apk, YMMV
     - During setup of the app, give access to the local device storage.
     - The app will create a folder `/devicestorage/vevs/`. Add a folder `logs` in this folder (`/devicestorage/vevs/logs/`).
     - Add the sensor device to the MiHome app.


Ernst79:
@jeremysherriff Thanks for your input. However, I used the latest version yesterday (6.0.214), and it writes pairings.txt just fine. However, I saw that someone mentioned that you had to make the `logs` folder yourself beforehand. But that's written in the instructions. Perhaps something like that went wrong. Perhaps, you could try it again.

jeremysherriff:
@Ernst79 yes I created the `logs` subdirectory first, the app the correctly created a structure under that (devices, misc etc) when I paired my first device. However no files (literally, no log files of any sort) were created when pairing the devices.

I removed that version and tried with an early 5.6.x and this time files were created, and so I attempted a couple of other versions assuming that the 6.x series was no longer able to achieve what we were needing.

I'll try with a modded v6 again and report findings.

Ernst79:
Thats strange, I had no issue at all yesterday. Let me know, otherwise, I will add a note in the FAQ.

jeremysherriff:
@Ernst79 ok I have tested, and it works fine.  The only difference I can think of is that this time I used a modern Android device (Samsung Galaxy S9+ phone, Android 10) whereas I originally tried on an old device (so that I did not disrupt my "main" Mi Home app) that is running Android 7.0 Nougat, and probably does not have the required Android API version to support the app. Normally the Play Store would ensure compatibility but side-loading the APK would have allowed me to bypass that, and as the app didn't actually crash I assumed it was working OK.

I will close this PR as unnecessary, but perhaps a note in the faq regarding a decent version of Android is appropriate as I can imagine that others will also use old or spare devices to avoid "disruption" of their primary Mi Home system by using a modded APK.  I can confirm that Android 7 and APK version 5.8 and 5.9 appear to work fine, it's just v6.x of the APK that doesn't work with the old device.

Ernst79:
Thanks. I also used an old android 6, and it worked fine with 6.0.214. But I will add a note.

jeremysherriff:
OK, that was a guess. Maybe it's to do with it being a tablet instead of a phone?  Whatever the reason, I've confirmed v6.x doesn't create the files on that device, but it looks like any recent version of v5.x does.  All versions work fine on my phone.

Ernst79:
I've added a note to the FAQ, to try an old version, if pairings.txt isn't created.

Issue #180, opened 2020-12-02
Add support for mjyd02yl-a nightlight


skynetua:
I would like to request support for mjyd02yl-a nightlight.
It provides motion, illuminance data, advertisements are encrypted  
[logs.txt](https://github.com/custom-components/ble_monitor/files/5628208/logs.txt)

Magalex2x14:
Hello.
Since the data is encrypted, I need an encryption key to look at it.
Or we can try it "blindly" - tell me which version of the component you are using, I will make changes to it, and we will see how it works.

skynetua:
I'm using the latest one 0.8.4
I can give you a key if it's more preferred

Magalex2x14:
Let's try first. I think we have all the required entity types. Later I will give you a link to the modified code.

Magalex2x14:
I was mistaken in saying that there are all types of entities for this sensor... As far as I understand, there will be time since the last movement and illuminance.
The contents of the const.py file should be replaced with the following content:
[const.py](https://github.com/custom-components/ble_monitor/files/5631006/const.py.txt)

Leave the `report_unknown` option enabled.
Do not forget to register the encryption key in the configuration.
Play around with the sensor and drop the contents of the log here (it will already be decrypted).
Pay attention to the entities, created for this sensor (battery, illuminance, light) - see if they change and how.

skynetua:
Three sensors are created but their state doesn't get any updates. It is unknown all the time. 
Adv data is not shown in logs anymore as device is not unknown now, I suppose.
[logs.txt](https://github.com/custom-components/ble_monitor/files/5635469/logs.txt)

Magalex2x14:
This is strange, since the component writes to the log in addition to unknown sensors, also unknown data from known sensors. Okay, I suggest not to figure out what's the matter now. Just give me the key, and I myself will decipher the data you sent earlier. You can throw the key here and change it, or you can send it to the mail (magalex28@gmail.com).

skynetua:
I sent the key and new logs to your email.

Magalex2x14:
@skynetua Thanks, received.

In addition, thanks to you, I found a serious data decryption bug in ExtendedAdvertisement packets (you probably have an Intel NUC or another platform with a modern bt-adapter).

Please wait a little while I rollover updates for all active branches...

Magalex2x14:
[...]

PR #245, opened 2021-01-24
MJYD02YL add motion sensor

Ernst79:
This PR adds support for the motion sensor of MJYD02YL sensors. 

It only uses the messages with ctrl frame `0F 00`. There are also messages with ctrl frame `17 10`, but these seem to be parsed not correct. Will look into that for a future release. See also #180


Ernst79:
0.9.11-beta has been released, which is an improved version of this PR. Please keep this PR open as a reminder, but do not merge (I will merge the 0.9.11-beta branch in stead)

PR #252, opened 2021-01-31
Add option to disable jagged humidity values for some sensors

jan-be:
The LYWSD03MMC and MHO-C401 sensors by default don't show any decimal places for the humidity value with this integration.

This makes some sense because they are only about 1% precise. However this is also causing some issues, especially when trying to build graphs from the data. Take a look at this graph:

Screenshot Grafana humidity

I think it is quite obvious, that the second half of it past 23:00 is preferred.

But because I guess some people prefer the other way, I have made it optional.


Ernst79:
Thanks, I'll look into it.

Ernst79:
It's not entirely clear what this jagged function does. If it's true, it converts a float to an integer? But I guess it's already a integer? 

```
        if self._jagged is True:
            self._measurements.append(int(data[self._measurement]))
        else:
            self._measurements.append(data[self._measurement])
```

jan-be:
No it is actually a float in the format e.g. 50.01 for 50.01% humidity. I guess it was added because the sensor isn't really that precise and therefore subject to noise. And that means the value is going go up and down slightly, even if the true humidity stays perfectly constant.

Ernst79:
Ah, ok. I will add your code in the upcoming beta 0.9.11. 

This reminds me that for LYWSD03MMC with custom firmware, we do have humidity in one decimal, I need to think if it is needed to change something.

Magalex2x14:
Hi guys. Here is a link to an illustration of the problem for which I had to discard the decimal part of the humidity ("jagged" humidity workaround): https://github.com/custom-components/ble_monitor/issues/7#issuecomment-595948254
The decimal part was like some kind of counter, increasing by 0.1 with each next packet. I checked the raw data before implementing this crutch.
@jan-be from which sensor and which firmware version did you receive your graph from? Perhaps there was a firmware update that resolved this issue? Then it may be worth removing this workaround altogether, and advising users who encounter this problem to update the firmware?

P.S. I currently have no way to check my sensors

Magalex2x14:
Follow a few messages from the link that I gave above - there is a more clear illustration of the problem- https://github.com/custom-components/ble_monitor/issues/7#issuecomment-595993860

Ernst79:
I don't see this behavior with my LYWSD03MMC (the missing data is me experimenting with ble_monitor)

image

Magalex2x14:
If I understood the situation correctly, @jan-be wants to add the ability to disable the "jagged" humidity workaround, since he does not observe a problem on his sensor. This workaround was in the component code almost from the very beginning of LYWSD03MMC support. To test your sensor, you need to disable it. See line 517 in the current sensor.py of the master branch.

Ernst79:
yeah, you're right, I'll try right now

jan-be:
> @jan-be from which sensor and which firmware version did you receive your graph from? Perhaps there was a firmware update that resolved this issue? Then it may be worth removing this workaround altogether, and advising users who encounter this problem to update the firmware?

I am using the custom firmware by pvvx

Ernst79:
aha, that explains it. pvvx (ATC firmware is measuring in 0,01 % for Humidity. 

https://github.com/pvvx/ATC_MiThermometer

I think it is better to make a different workaround than. I made a device attribute called firmware (in the next version), which tells me that it is using ATC firmware or Xiaomi firmware. We can use that to turn the jagged option to false. We don't need a setting for that.

jan-be:
Ah I see, that's even better. I guess I am going to close the pull request then.

Ernst79:
I will change it like this (but won't work now, as `self._device_firmware` is new in the upcoming release. 

```
        # LYWSD03MMC / MHO-C401 "jagged" humidity workaround
        if devtype in ('LYWSD03MMC', 'MHO-C401'):
            if self._device_firmware is "Xiaomi (MiBeacon)"
                self._jagged = True
```

Anyway, @Magalex question is still relevant, it looks like your sensor isn't suffering from this strange behavior.

Magalex2x14:
My sensors have original firmware. As soon as I have the opportunity, I will disable the workround and test the behavior with the original firmware. If my memory serves me, there has been a firmware update since the implementation of support in our integration. Custom firmware can calculate data from the sensor chip differently, that is, it does not contain this problem "out of the box".

Ernst79:
0.9.11-beta has been released, which adds the above described fix. It now only uses the jagged humidity workaround for Xiaomi firmware, not for ATC firmware.

PR #291, opened 2021-02-24
Improvement in errors to show in logs.

SeiTaN:
Hi there,

To improve error messages I think is better handling them instead return a None, now when an error is raisen you'll have an entry in the log section.

What do you think about it?

Tested in Xiaomi and ATC devices.


Ernst79:
Thanks, I'm testing it right now. seems to be working fine (although I didn't have an error yet). I'll merge it in the next version and will look at the TODO. The first one isn't an error, just a first value in case it has just started without a previous message.

SeiTaN:
You are welcome, you may test the funcionality just adding `raise NoValidError("Custom error")` at the beginning of any parse method.
To clarify, you can get a lot of error messages because the component's scan loop is around 5 seconds (in my tests).

Issue #233, opened 2021-01-12
Bug - struct.error: unpack requires a buffer of 4 bytes


perseus177:
Environment 

Passive BLE monitor integration (Xiaomi Mijia BLE MiBeacon monitor)
Installed version: 0.9.4 
------------------------------------------------------------------------------------
Version | 2021.1.0
-- | --
Installation Type | Home Assistant Supervised
Development | false
Supervisor | true
Docker | true
Virtual Environment | false
Python Version | 3.8.7
Operating System Family | Linux
Operating System Version | 5.4.72-v7+
CPU Architecture | armv7l
------------------------------------------------------------------------------------

Logger: homeassistant
Source: custom_components/ble_monitor/__init__.py:366
First occurred: 10. januára 2021, 16:57:47 (1 occurrences)
Last logged: 10. januára 2021, 16:57:47
Error doing job: Fatal error: protocol.data_received() call failed.

Traceback (most recent call last):
`  File "/usr/local/lib/python3.8/asyncio/selector_events.py", line 862, in _read_ready__data_received
    self._protocol.data_received(data)
  File "/config/custom_components/ble_monitor/aioblescan_ext.py", line 1497, in data_received
    self.process(packet)
  File "/config/custom_components/ble_monitor/__init__.py", line 483, in process_hci_events
    msg, binary, measuring = self.parse_raw_message(data)
  File "/config/custom_components/ble_monitor/__init__.py", line 736, in parse_raw_message
    result.update(resfunc(xvalue))
  File "/config/custom_components/ble_monitor/__init__.py", line 366, in obj0d10
    (temp, humi) = TH_STRUCT.unpack(xobj)
struct.error: unpack requires a buffer of 4 bytes`

Ernst79:
Hi, thanks for reporting. 

This message indicates that the BLE message is recognized as Xiaomi Mi_Beacon message, 0d10 means that the message should contain temperature and humidity data in a 4 bytes long data message (2 for temperature, 2 for humidity), but apparently, the BLE data message is not 4 bytes long. 

So, to figure out what is wrong, first a couple of questions, what sensor do you have?
And two, is this error an incident (happens once) or does it occur more often?

perseus177:
Hello 

I have 

- 3 sensors LYWSDCGQ (Temperature and Humidity)
- 1 Kettle YM-K1501

Error occurred twice from yesterday.

Ernst79:
[...]

PR #294, opened 2021-02-26
revert NoValidError for sensor not in whitelist

### LINKED ISSUE #293: Invalid data: MAC is not present in whitelist
After updating to version 1.1.2 an error occurred:

Logger: custom_components.ble_monitor
Source: custom_components/ble_monitor/__init__.py:673
Integration: Passive BLE monitor (documentation)
First occurred: 10:10:48 (149 occurrences)
Last logged: 10:13:11

Invalid data: MAC is not present in whitelist

Ernst79:
Reverted the error that was raised when a device is not in the whitelist and discovery is set to false. 

@SeiTaN This is not an error, but an intentional configuration by the user. So I reverted the error handling for this specific occasion. 

Fixes #293

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -722,7 +722,7 @@ class HCIdump(Thread):
 
         # check for MAC presence in whitelist, if needed
         if self.discovery is False and xiaomi_mac_reversed not in self.whitelist:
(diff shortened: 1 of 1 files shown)


Ernst79:
Added fix for #233 This sensor LYWSDCGQ seems to send occasional messages with 5 bytes. It now checks for the length and if not 4 bytes, it skips the message.

PR #314, opened 2021-03-29
Support for adapter address instead of interface number

daniele-athome:
This patch prevents issues when Linux assigns random number to hci interfaces (I have 2 USB adapters and hci interface number is assigned randomly at boot). I couldn't find a way to persist the hci interface number in any way (please advise if you know how to do it), so I decided to query the system bluetooth library for the device interface number.

This is somewhat a draft that I won't complete. Things to do:

* support for FreeBSD - I don't really know how it works
* introducing a dependency on the bluetooth shared system library that can't be declared in manifest.json (can it?)
* extensive tests, especially for backwards compatibility (although device interface number should be completely replaced by adapter address IMHO)

I'm not providing a complete patch because of lack of time but I thought I'd share what I've done so far and let others continue the development, if interested.

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -4,6 +4,7 @@ import copy
 import json
 import logging
 import math
+import platform
 import struct
 from threading import Thread
 import janus
@@ -128,7 +129,7 @@ CONFIG_SCHEMA = vol.Schema(
                     vol.Optional(CONF_ACTIVE_SCAN, default=DEFAULT_ACTIVE_SCAN): cv.boolean,
                     vol.Optional(
                         CONF_HCI_INTERFACE, default=[DEFAULT_HCI_INTERFACE]
-                    ): vol.All(cv.ensure_list, [cv.positive_int]),
+                    ): vol.All(cv.ensure_list, [cv.string]),
                     vol.Optional(
                         CONF_BATT_ENTITIES, default=DEFAULT_BATT_ENTITIES
                     ): cv.boolean,
@@ -148,6 +149,21 @@ CONFIG_SCHEMA = vol.Schema(
 )
 
 SERVICE_CLEANUP_ENTRIES_SCHEMA = vol.Schema({})
+
+
+if platform.system() == "Linux":
+    from ctypes import CDLL
+    from ctypes.util import find_library
+
+    btlib = find_library("bluetooth")
+    if not btlib:
+        raise Exception(
+            "Can't find required bluetooth libraries"
+            " (need to install bluez)"
+        )
+    bluez = CDLL(btlib, use_errno=True)
+else:
+    bluez = None
 
 
 async def async_setup(hass: HomeAssistant, config):
@@ -247,7 +263,7 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry):
     else:
         hci_list = config_entry.options.get(CONF_HCI_INTERFACE)
         for i, hci in enumerate(hci_list):
-            hci_list[i] = int(hci)
+            hci_list[i] = str(hci)
         config[CONF_HCI_INTERFACE] = hci_list
     _LOGGER.debug("HCI interface is %s", config[CONF_HCI_INTERFACE])
 
@@ -627,11 +643,15 @@ class HCIdump(Thread):
             if self._event_loop is None:
                 self._event_loop = asyncio.new_event_loop()
                 asyncio.set_event_loop(self._event_loop)
-            for hci in self._interfaces:
+            for hci, addr in enumerate(self._interfaces):
                 try:
-                    mysocket[hci] = aiobs.create_bt_socket(hci)
+                    if type(addr) == int or addr.isdigit():
+                        dev_id = int(addr)
+                    else:
+                        dev_id = bluez.hci_get_route(addr)
+                    mysocket[hci] = aiobs.create_bt_socket(dev_id)
                 except OSError as error:
-                    _LOGGER.error("HCIdump thread: OS error (hci%i): %s", hci, error)
+                    _LOGGER.error("HCIdump thread: OS error (hci%i): %s", addr, error)
                 else:
                     fac[hci] = getattr(self._event_loop, "_create_connection_transport")(
                         mysocket[hci], aiobs.BLEScanRequester, None, None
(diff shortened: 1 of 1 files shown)


Ernst79:
Thanks, I think I can implement this some way. Have to think about it what’s the best place, can also be implemented in aioblescan_exp.py

daniele-athome:
I noticed there are a few errors in the code (it works but errors are not handled well)... sorry I did this in a rush :-) but you get the idea.

Ernst79:
No problem, I get the idea.

Ernst79:
Started to work on it. Added automatic MAC discovery in UI mode. 

image

daniele-athome:
Spending a little more time on this... I found out that the shared library method doesn't really work. It's way too low level anyway.
This should be the correct approach:

```
import bluetooth
dev_id = bluetooth._bluetooth.hci_devid(addr)
```

This will correctly return 0, 1, 2, ... matching hci0, hci1, hci2 and so on. And it requires a dependency to `pybluez`, but I guess it's better than the shared library :-)

Sorry again for doing this in a rush.

Ernst79:
I use this now to get a dictionary which links hci to the MAC address.

```
def get_bt_interface_mac(interface_list=[0]):
    # Get dict of available bluetooth interfaces, returns hci and mac
    btaddress_dict = {}
    status, output = subprocess.getstatusoutput("hciconfig")
    
    for interface in interface_list:
        hci_id = "hci{}".format(interface)
        try:
            btaddress_dict[interface] = output.split("{}:".format(hci_id))[1].split("BD Address: ")[1].split(" ")[0].strip()
        except:
            pass
    return btaddress_dict
```

I have it working now, but have to do some testing.

daniele-athome:
Oh, ok. But note that it's a dependency to an executable that you can't put anywhere (in a declarative way I mean).

Ernst79:
The problem is that it is quite difficult to get the Mac addresses on home assistant OS yourself, as you don’t have access to the command line normally. So we need to have something that helps a user to get the MAC address.

In what situation would the above code not work? 

It works fine on home assistant OS right now. IBut if there are better ways, I’m all ears..

daniele-athome:
You're right sorry I was accounting only for a virtualenv installation. So the HA container has hciconfig, I didn't know.

I don't have any ideas other than use pybluez (a library). What does HA uses for bluetooth anyway?

Ernst79:
> What does HA uses for bluetooth anyway?

HA uses python3, so whatever works in python, works in HA. Each integration is using it's own bluetooth implementation, but I guess the two main implementations are pybluez or native Python sockets support for Bluetooth in python, which is added in python 3.3. 

https://blog.kevindoran.co/bluetooth-programming-with-python-3/ 

In the file aioblescan_ext.py in our integration, they use `socket`.

```
        sock = socket.socket(
            family=socket.AF_BLUETOOTH, type=socket.SOCK_RAW, proto=socket.BTPROTO_HCI
        )
```

According to the documentation, there are 4 ways to use this, 

> AF_BLUETOOTH supports the following protocols and address formats:
> 
> - BTPROTO_L2CAP accepts (bdaddr, psm) where bdaddr is the Bluetooth address as a string and psm is an integer.
> 
> - BTPROTO_RFCOMM accepts (bdaddr, channel) where bdaddr is the Bluetooth address as a string and channel is an integer.
> 
> - BTPROTO_HCI accepts (device_id,) where device_id is either an integer or a string with the Bluetooth address of the interface. (This depends on your OS; NetBSD and DragonFlyBSD expect a Bluetooth address while everything else expects an integer.)
> 
> - BTPROTO_SCO accepts bdaddr where bdaddr is a bytes object containing the Bluetooth address in a string format. (ex. b'12:23:34:45:56:67') This protocol is not supported under FreeBSD.

So, we now use the BTPROTO_HCI, but we can also use the BTPROTO_SCO option if we want to use the MAC directly. 

However, to keep things backwards compatible, I was thinking to allow a user to use either `hci_interface: 0` or `bt_interface: AA:BB:EE:11:22:33` in yaml configuration (YAML mode), and than to convert the mac address to the hci number in python. The hciconfig command allows me to create a dictionary with the relation between hci and mac, but also allows me to create a list with available macs, which can be shown in the user interface (UI mode) as a drop down list in the options. It also allows us to use the first available BT interface as default, in stead of hci0, which might not exist at all. 

I have it working when using UI mode, but have to test in YAML mode. I also want to implement an automatic conversion from HCI to MAC for users that use UI mode. In UI mode, you can only select an MAC (no HCI), in yaml mode, you can use both.

daniele-athome:
Thanks for the detailed explanation.
 
> However, to keep things backwards compatible, I was thinking to allow a user to use either hci_interface: 0 or bt_interface: AA:BB:EE:11:22:33 in yaml configuration (YAML mode), and than to convert the mac address to the hci number in python.

It seems fair (and useful in general).

> I have it working when using UI mode, but have to test in YAML mode. I also want to implement an automatic conversion from HCI to MAC for users that use UI mode. In UI mode, you can only select an MAC (no HCI), in yaml mode, you can use both.

I'll be eagerly waiting for something to test :-)

Ernst79:
> I'll be eagerly waiting for something to test :-)

I'll do the testing tonight. Hopefully I can upload the first draft tonight. Are you using YAML of UI configuration?

daniele-athome:
YAML.

Ernst79:
Sorry, need a bit more time. Had to apply some fixes to YAML mode to make it backwards compatible. I hope to finish after Eastern.

daniele-athome:
No problem man, take your time.

Ernst79:
It's finished, I think.

You can try by copying all updated files from [this branch](https://github.com/custom-components/ble_monitor/tree/bt_interface-via-mac/custom_components/ble_monitor)

Configuration in YAML is also described here.
https://github.com/custom-components/ble_monitor/blob/bt_interface-via-mac/README.md#bt_interface

It should be backwards compatible. UI mode will change to MAC only (automatic conversion), YAML mode will keep both options, to keep backwards compatibility. Note that `hci_interface` will overrule `bt_interface`

Could you give it a try?

daniele-athome:
Thank you! I'll give it a try tonight.

daniele-athome:
I tested these use cases:

- requested adapter was found
- requested adapter was not found
- requested adapter was plugged in after HA start

When the requested adapter was not found, the component falled back gracefully to the default one:

```
WARNING (MainThread) [custom_components.ble_monitor] No configured Bluetooth interfaces was found, using default interface instead
```

The third case didn't work probably because HA doesn't really reload classes when you ask it to reload an integration (I see BT_INTERFACES is global, so it's probably evaluated once). But that's fine, it's a very niche use case, restarting HA fixes the problem.

Thank you!!! I guess you may close this PR which became more a feature request than a real contribution eheh.

Ernst79:
Thanks for testing. I’ll create a beta tomorrow.

the third usecase is indeed not covered, but, as you say, a niche. 

I’ll close this PR and make a new one tomorrow. Thanks for the idea and testing :-)

Ernst79:
PR is in #318

PR #338, opened 2021-04-30
Add Quingping CGG1 variant

swingerman:
Added another device type header for my CGG1's. And magically they showed up after a year of struggle.

My devices have no Quingping logo on the back and they don't work with Mi Home app. They only work with the Quingping app.

FCCID: 2AQ3F-CGG1


Ernst79:
Thanks, I'll add it in the next release.

Ernst79:
Added in 1.9.1-beta

PR #363, opened 2021-05-24
Fix log message level

tomtoump:
The component uses the`warning` level for listing the available Bluetooth interfaces, instead of `info`.

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -239,7 +239,7 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry):
         # Configuration in YAML
         for key, value in CONFIG_YAML.items():
             config[key] = value
-        _LOGGER.warning("Available Bluetooth interfaces for BLE monitor: %s", BT_MAC_INTERFACES)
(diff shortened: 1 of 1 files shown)


Ernst79:
Thanks, but the problem with info level is that the interfaces to select from are not shown with default configuration. I did this such that a user can easily see the MAC address, and use that in his YAML configuration. 

When setting it to info level, a user wont be able to see the MAC address easily (he will have to enable logger at info level, restart, etc). But a workaround might be program it such that it only show this message if it not configured yet.

tomtoump:
Your solution works as well in order to stop polluting the logs when there is no need to. I also removed the default value for `CONF_BT_INTERFACE` since [this](https://github.com/custom-components/ble_monitor/blob/f50b4e420145fa1c67bb1ddfd0ea2bb3bb9ea85c/custom_components/ble_monitor/__init__.py#L267) condition never triggered.

Ernst79:
I think this condition triggers when you enter a wrong MAC in your yaml config. 

After a second thought, I think your first solution is actually OK (sorry). Most people won’t specify it anyway, and just use the default. I’ll add instructions in the documentation for those who do want to specify it. I’ll add it in the next release. Thanks.

Ernst79:
Your first solution is merged into the Master branch. I will close this one. Thanks

PR #367, opened 2021-05-27
Update and rename en.json to zh_TW.json

bluefoxlee:


Ernst79:
Do not delete `en.json`, only add your own language file.

bluefoxlee:
Forget to duplicate instead of rename. My bad.

PR #369, opened 2021-05-27
Create zh-Hant.json

bluefoxlee:
Made few mistakes on adding zh-Hant.json.


bluefoxlee:
> Thanks, I'll merge this one. Can't check it unfortunately, in the Netherlands they don't really educate the Chinese language.
> 
> It will be included in the next (final) release.

will do proof reading couple time next time before I submit. It’s hard to catch minor issue when doing the translation and proof read at same time.

PR #425, opened 2021-07-04
Add device tracker and iNode energy meter support

Ernst79:
This PR adds support for device trackers
- 3 options added
  - `track_device: True` to enable tracking
  - `tracker_scan_interval: 20` to reduce the update interval in HA
  - `consider_home: 120` to set the time after which the device tracker is going back to `Away`
- `restore_state`, `mac` and `name` should work similar as for the sensors. 

config in YAML

```YAML
ble_monitor:
  devices:
    - mac: 'D4:3C:2D:4A:3C:D5'
      track_device: True
      tracker_scan_interval: 20
      consider_home: 180
```

## Other updates in this PR

- This PR also adds support for iNode Energy Meter
- This PR also adds support for the Dutch language


Magalex2x14:
Just for clarification I would add that the tracker will only work with devices using a public or random static (lifetime) address. You can read more about address types [here](https://www.novelbits.io/bluetooth-address-privacy-ble/), for example. By the way, in ESPHome, in addition to using the address, it is possible to use a fixed UUID for tracking (advertisement of a certain type with a fixed payload) - it is quite possible that the release will be followed by a corresponding user request...

Ernst79:
Thanks, I was expecting that such questions would pop up. But I wanted to make a start with something, expending functionality it easier when we have a start. Will add your suggestion to the docs tomorrow.

Issue #352, opened 2021-05-19
Yeelight remotes


Ernst79:
Request from @bluefoxlee from #289 

Hi.
Wonder is there any possible to add support for this remote since it's very similar to 01YL but slightly different.
It comes with Yeelight celling fan/light product. Got PID 1678 through Miiocli. however can't find the model number of it.

![image](https://user-images.githubusercontent.com/24375847/118835135-03fd5680-b8c3-11eb-9d76-910b8c3364ee.png)
[Left: YLYK01YL] [Right: PID 1678]

Ernst79:
@bluefoxlee can you log it with hcidump or ble_monitor debug log? each button single click, double click and long press. pretty sure it can be supported.

```
  var RemoteControllerType = {
    BLERC: 339,
    SEESAW: 950,
    BHFRC: 959,
    VENFAN: 1254,
    FANCL: 1678
  };
```

Or 

The easiest to try is modify the following two files, In each file look for YLYK01YL, copy this line and modify to YLYKxxYL). In the first file, also modify b'\x53\x01 to b'\x8E\x06 in the new line.

`\custom_components\ble_monitor\ble_parser\xiaomi.py`

```
    b'\x53\x01': ("YLYK01YL", True),
    b'\x8E\x06': ("YLYKxxYL", True),
```

and in `\custom_components\ble_monitor\const.py`

```
    'YLYK01YL'       : [[9, 9, 9, 9, 9, 9, 9, 9, 9, 0, 9, 9, 9, 9, 9, 9, 9], [0, 9, 9, 9, 9, 9, 9, 9]],
    'YLYKxxYL'       : [[9, 9, 9, 9, 9, 9, 9, 9, 9, 0, 9, 9, 9, 9, 9, 9, 9], [0, 9, 9, 9, 9, 9, 9, 9]],
```

and

```
    'YLYK01YL'       : 'Yeelight',
    'YLYKxxYL'       : 'Yeelight',
```

For now, states will probably the same as for the YLYK01YL remote, but let first find out if this works. You might need an encryption key (24 characters). You can follow the instructions here, for 'ceiling light', you can read 'fan', I guess (seems like a fan remote).

I think the remote belongs to the YLFD001 ceiling fan light, isn't there a name on the back or insight an battery compartement?

bluefoxlee:
Thanks @rezmus and @Ernst79 for replay and help to open an issue. I will make some test with it and report it back tomoroww.

[...]

PR #469, opened 2021-08-13
fix(xiaomi parser): change fan_remote_control from min to max

Thrilleratplay:
When looking up broadcasted properties, I noticed `fan_remote_command = "brightness min"` was set for two buttons.  I am assuming for + it should be `fan_remote_command = "brightness max".

### DIFF
--- a/custom_components/ble_monitor/manifest.json
+++ b/custom_components/ble_monitor/manifest.json
@@ -14,6 +14,6 @@
     "@Ernst79",
     "@Magalex2x14"
   ],
-  "version": "4.5.0-beta",
+  "version": "4.5.0",
   "iot_class": "local_polling"
 }


custom_components/ble_monitor/ble_parser/xiaomi.py
```diff
@@ -563,14 +586,18 @@ def parse_xiaomi(self, data, source_mac, rssi):
             if payload_length < next_start:
                 _LOGGER.debug("Invalid payload data length, payload: %s", payload.hex())
                 break
-            object = payload[payload_start + 3:next_start]
+            object = payload[payload_start + 3 : next_start]
```
Ernst79, line 589: I think the spaced around : should be removed, I’m unable to test it right now if this also works, normally the format is payload[3:9], isn’t it?

Ernst79 (changes requested):

Ernst79:
I think it should have been 3 = color temperature, 5 = brightness, see the FANCL picture

custom_components/ble_monitor/ble_parser/xiaomi.py
Thrilleratplay: Crap.  I didn't notice Python Black ran on save.  I only was purposing the change on line 142.  I'll revert everything else.

Thrilleratplay:
@Ernst79 I am trying to find information on this version of the remote and pretty much everything related to it is from this repo.  What an oddity and why put those functions on a rocker button? Anyways, [based on the comment from the owner](https://github.com/custom-components/ble_monitor/issues/352#issuecomment-845715963) you are correct.  I'll change both values without Python Black's reformatting later today.

Issue #467, opened 2021-08-12
govee 5051 connected, but not passing temp/humidity date


fancybeachsnake:
in response to a q/a session with @Thrilleratplay, was advised to open an issue on this.

have a govee 5051, device appears to be connected, but is not providing temp/humidity data.

home-assistant.log with ble_monitor in debug, and hcidump attached.

[home-assistant.log](https://github.com/custom-components/ble_monitor/files/6978801/home-assistant.log)
[hcidump.txt](https://github.com/custom-components/ble_monitor/files/6978802/hcidump.txt)

Ernst79:
Could you try it with temperature set to Celsius in the BLE monitor options?. I think the sensor sends temperature data in C, HA will convert it to F if your user setting is F in HA. Setting it to F is only needed if the sensor is sending data in F.

Ernst79:
@Thrilleratplay I see messages in the HCI dump with length 13 (0C) with FF 88 EC in stead of 11 (0A). We might have to use these, but I’m unable to look into that coming 1.5 week. Perhaps you can have a look, otherwise I can do it later

Thrilleratplay:
@fancybeachsnake @Ernst79 I tried looking into this last night and noticed the size dependency but after repeatedly confusing myself trying to map this along with H5074 information, I called it a night.  I'll take a look later today.

Thrilleratplay:
@fancybeachsnake just to make sure I am parsing the correct packet, when you created hcidump.txt, would the values for temperature 27.4C/81.F and humidity 40% seem correct?

fancybeachsnake:
> Could you try it with temperature set to Celsius in the BLE monitor options?. I think the sensor sends temperature data in C, HA will convert it to F if your user setting is F in HA. Setting it to F is only needed if the sensor is sending data in F.
> 
> I also see 'temperature_unit': 'Â°F', in the log, this should be 'temperature_unit': 'F' or ‘C’, without the degree symbol and without the Â
> 
> and is the Mac of the sensor in the devices option? You have turned of discovery, so you will have to define the Mac.

reset it to C.

the mac addr was in the devices option this whole time, but re-enabled discovery.

no change in attributes being returned/parsed, at least from ha's perspective.  i am unclear how to read the output from hcidump, i presume it's hex.

the weird formatting in from of the 'F' i believe is strange encoding from copy/paste on a mac, my local log does not reflect that character.

fancybeachsnake:
[...]

PR #471, opened 2021-08-13
Fix for Govee H5051

Thrilleratplay:
I believe this will fix the issue in #467 but am unable to test as I do not own one.  There is another change that adds mapping from one device model name to another to prevent breaking the component for existing users when splitting H5051 and H5074.  This is not the most elegant solution but am open to ideas.


Ernst79:
Changing device names is a bit problematic yes, I have never found a way to change the device name in the registry. Perhaps this is fine for now, but I guess the best solution should be renaming the device name in the device info. But I’m not sure that is even possible.

Ernst79 (approved):

Ernst79:
I’ll merge it for now and will release a beta. Maybe later, if I ever find how to update the device info, we can modify this.

Thrilleratplay:
@Ernst79 I did a diff of `core.device_registry` before and after, the new device names to appear to be saved/updated but could not find in `sensor.py` where this occurs.  While this is a bit of a hack, it does seem to actually update the device's model property.

Ernst79:
Ok, that’s great. Can’t test it myself at the moment, enjoying summer holidays.

PR #480, opened 2021-08-20
Initial version of SensorPush integration

relevante:
Hello! We've had a number of customers express interest in using our devices directly via Bluetooth with Home Assistant. One of them pointed us to this project, and I noticed in the issues log that someone had tried to implement it awhile back, but didn't have access to the protocol details and got stuck.

I put together a parser and tests, following the patterns of the others. It works with our HT.w and HTP.xw sensors.

I've been testing it on my local install and one of our customers has as well, with good results thus far.

Happy to contribute to this very cool project. Hope people find the addition valuable!


Ernst79:
Thank you very much for your contribution, much appreciated. I will look into it a.s.a.p., probably after the weekend.

Ernst79:
It has been released as version 5.0.0. We also updated our documentation to a Jekyll documentation, so I had to rewrite some things of the SensorPush sensors to fit in the devices placeholder format. 

The device info goes into these files 
https://github.com/custom-components/ble_monitor/blob/master/docs/_devices/SensorPush_HTP_xw.md
https://github.com/custom-components/ble_monitor/blob/master/docs/_devices/SensorPush_HT_w.md

To view the final result in the Jekyll documentation
https://custom-components.github.io/ble_monitor/by_brand#sensorpush

If you want, you can make a final check. 

Thanks again for your contribution

melyux:
@relevante I found that this parser makes use of extremely large numbers for modulo (it uses 273613520207001 to calculate pressure). Is there a way to do this on a 32-bit system?

Ernst79:
I have no idea what you mean? Can you open a new issue and explain a bit further?

melyux:
I commented here because @relevante  wrote the only decoder I've ever seen for these sensors, right here.

This decoder works by using extremely large modulo operands. Running it on an embedded system, it overflows and gives garbage data because of this. Was looking for a way around that

relevante:
> This algorithm works by using extremely large modulo operands. Running it on an embedded system, it overflows and gives garbage data because of this. Was looking for a way around that

You should be able to make it work on virtually any system. Even on 32-bit systems, compilers generally support 64-bit variables and math operations.

Essentially the same code you see in the parser runs on the 32-bit microcontroller on the sensor. The variables just have to be declared as uint64_t, and we declare the floating point values as doubles. The C compiler handles the rest.

PR #485, opened 2021-08-29
Add SensorEntityDescription to BLE monitor

Ernst79:
This PR adds the new SensorEntityDescription and BinarySensorEntityDescription, which allows us to significantly reduce the code of the sensor.py and binary_sensor.py file. 

I also have fixed some style formatting in the Home Assistant development environment, which is mostly cosmetic. 

This version might contain bugs, as it was a significant rewrite. In the future, I will do the same for the binary sensors. 

BREAKING CHANGES
- ext_state attribute has been renamed to status and is displaying the actual meaning in stead of a number


Ernst79:
Test fail on the new "native_unit_of_measurement". I'll wait a short while till `pytest-homeassistant-custom-component` has been updated to HA 2021.9

Issue #441, opened 2021-07-15
Add support for Moat S2


Thrilleratplay:
With the inclusion of Govee devices, I wanted to highlight the work of @SteveOnorato who created a component based on [Home-Is-Where-You-Hang-Your-Hack/sensor.goveetemp_bt_hci](https://github.com/Home-Is-Where-You-Hang-Your-Hack/sensor.goveetemp_bt_hci) that includes a similar device, the Moat S2 BLE Temperature & Humidity sensor.  

The repo is: [SteveOnorato/moat_temp_hum_ble](https://github.com/SteveOnorato/moat_temp_hum_ble)

Ernst79:
Added in 5.3.0. If anyone can confirm the correct working, that would be appreciated.

PR #506, opened 2021-09-10
Add support for Moat S2

Ernst79:
Added support for Moat S2. I was not able to confirm that it is working properly, based the parser on https://github.com/SteveOnorato/moat_temp_hum_ble. Fix for #441


Ernst79:
@Thrilleratplay Could you do a check if you see something strange? I don't have this device (I guess you do not have it either).

Thrilleratplay:
@Ernst79 Sorry, I don't have one either.  The temperature and humidity looks correct based on [temp_hum_advertisement_parser.py](https://github.com/SteveOnorato/moat_temp_hum_ble/blob/master/custom_components/moat_temp_hum_ble/temp_hum_advertisement_parser.py) but never took a close look at the battery calculation.  Seeing that the voltage is returned as well, I suppose battery percentage can be adjusted later when more information is available.  

@SteveOnorato would you be able to test this when you the time?

Ernst79:
I copied the battery calculations from another sensor brand. It is different from the definition that @SteveOnorato is using. Perhaps he has calibrated it with an app? If that is the case, I can better use his formula. I’ll wait to see if @SteveOnorato is willing to test.

SteveOnorato:
> I copied the battery calculations from another sensor brand. It is different from the definition that @SteveOnorato is using. Perhaps he has calibrated it with an app? If that is the case, I can better use his formula. I’ll wait to see if @SteveOnorato is willing to test.

I'm not sure when I'll be able to get around to trying this with the actual devices, but I took a quick look at the code.  The battery % calculation in my code is my own placeholder estimate, and as the comment suggests, I was hoping to eventually update it with something more realistic: https://github.com/SteveOnorato/moat_temp_hum_ble/blob/master/custom_components/moat_temp_hum_ble/temp_hum_advertisement_parser.py#L67 

However, I'm still on my first set of batteries, so I still don't have an idea of how long they actually last.  I put two S2's into service in Sept 2020, and those year-old batteries are reporting 2953mV (at 47.9 deg F) and 2779mV (at 1.2 deg F) today.

So the calculations in your code are at least better than mine, but still would report 95.3% for one of my year-old batteries, which seems misleading.  I suspect the estimations at https://devzone.nordicsemi.com/f/nordic-q-a/6914/battery-level-discharge-curve-how-can-i-read are slightly better, but still inexact and disregard temperature & load:
 * Section 1: 3.0V - 2.9V = 100% - 42% (58% drop on 100 mV)
 * Section 2: 2.9V - 2.74V = 42% - 18% (24% drop on 160 mV)
 * Section 3: 2.74V - 2.44V = 18% - 6% (12% drop on 300 mV)
 * Section 4: 2.44V - 2.1V = 6% - 0% (6% drop on 340 mV)

The actual Moat S2 battery is a CR2450 (not sure which brand), but from this datasheet we can see just how much the temperature affects the voltage: https://biz.maxell.com/en/primary_batteries/CR2450HR_DataSheet_20e.pdf

Ernst79:
Ok, thanks. If you have a hcidump of a ble advertisement of a Moat S2, it would also be enough for me to verify the code. I had to reverse engineer a BLE message, based on your code, I think the biggest risk is that I made a mistake there.

Ernst79:
Battery is updated

Ernst79:
I’ll merge it for now, but I appreciate if someone can let me know if it all works ok

PR #540, opened 2021-10-31
Update zh-Hant.json

bluefoxlee:


Ernst79:
Thanks, will add it in the next update

PR #572, opened 2021-12-03
Added parse_data service

myhomeiot:
Send RAW HCI packet data to the BLE Montitor integration, used by ESPHome [BLE Gateway](https://github.com/myhomeiot/esphome-components#ble-gateway)

We discussed this [here](https://community.home-assistant.io/t/passive-ble-monitor-integration/303583/190?u=myhomeiot)

This code tested and fully working but I will be glad if you take a look into `async_parse_data_service` function and change it as you think it's should be.

Thanks in advance!

PS: You can also add [user permission check](https://developers.home-assistant.io/blog/2019/03/11/user-permissions) for service calls to prevent regular users calling services, by changing 

```
    hass.services.async_register(
        DOMAIN,
        SERVICE_CLEANUP_ENTRIES,
        service_cleanup_entries,
        schema=SERVICE_CLEANUP_ENTRIES_SCHEMA,
    )
```
to
```
    platform = entity_platform.async_get_current_platform()
    platform.async_register_entity_service(
        SERVICE_CLEANUP_ENTRIES,
        SERVICE_CLEANUP_ENTRIES_SCHEMA,
        "async_cleanup_entries_service"
    )
```

As example in my integration I did it in this [commit](https://github.com/myhomeiot/DahuaVTO/commit/d23268848921adf7d065d68ece234af82b9ff3b0)

### DIFF
--- a/custom_components/ble_monitor/__init__.py
+++ b/custom_components/ble_monitor/__init__.py
@@ -48,6 +48,7 @@ from .const import (
     CONF_REPORT_UNKNOWN,
     CONF_RESTORE_STATE,
     CONF_USE_MEDIAN,
+    CONF_PACKET,
     CONFIG_IS_FLOW,
     DEFAULT_ACTIVE_SCAN,
     DEFAULT_BATT_ENTITIES,
@@ -72,6 +73,7 @@ from .const import (
     MEASUREMENT_DICT,
     REPORT_UNKNOWN_LIST,
     SERVICE_CLEANUP_ENTRIES,
+    SERVICE_PARSE_DATA,
 )
 
 from .bt_helpers import (
@@ -167,6 +169,11 @@ CONFIG_SCHEMA = vol.Schema(
 )
 
 SERVICE_CLEANUP_ENTRIES_SCHEMA = vol.Schema({})
+SERVICE_PARSE_DATA_SCHEMA = vol.Schema(
+    {
+        vol.Required(CONF_PACKET): cv.string,
+    }
+)
 
 
 async def async_setup(hass: HomeAssistant, config):
@@ -177,11 +184,22 @@ async def async_setup(hass: HomeAssistant, config):
 
         await async_cleanup_entries_service(hass, service_data)
 
+    async def service_parse_data(service_call):
+        service_data = service_call.data
+
+        await async_parse_data_service(hass, service_data)
+
     hass.services.async_register(
         DOMAIN,
         SERVICE_CLEANUP_ENTRIES,
         service_cleanup_entries,
         schema=SERVICE_CLEANUP_ENTRIES_SCHEMA,
+    )
+    hass.services.async_register(
+        DOMAIN,
+        SERVICE_PARSE_DATA,
+        service_parse_data,
+        schema=SERVICE_PARSE_DATA_SCHEMA,
     )
 
     if DOMAIN not in config:
@@ -392,6 +410,13 @@ async def async_cleanup_entries_service(hass: HomeAssistant, service_data):
         if len(async_entries_for_device(entity_registry, device_id)) == 0:
             device_registry.async_remove_device(device_id)
             _LOGGER.debug("device %s will be deleted", device_id)
+
+
+async def async_parse_data_service(hass: HomeAssistant, service_data):
+    """Call parse_data with RAW HCI packet data."""
+    _LOGGER.debug("async_parse_data_service")
+    blemonitor = hass.data[DOMAIN]["blemonitor"]
+    blemonitor.dumpthread.process_hci_events(bytes.fromhex(service_data["packet"]))
 
 
 class BLEmonitor:
--- a/custom_components/ble_monitor/const.py
+++ b/custom_components/ble_monitor/const.py
@@ -65,9 +65,11 @@ CONF_DEVICE_TRACK = "track_device"
 CONF_DEVICE_TRACKER_SCAN_INTERVAL = "tracker_scan_interval"
 CONF_DEVICE_TRACKER_CONSIDER_HOME = "consider_home"
 CONF_DEVICE_DELETE_DEVICE = "delete device"
+CONF_PACKET = "packet"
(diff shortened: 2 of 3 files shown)


Ernst79:
Nice work, I will try to test is a.s.a.p. If I understand it correct, you use an ESPHome device to receive BLE messages and than broadcast the same data again? So it is working as some kind of "range extender". 

Just curious, what is the benefit of this, ESPHome is also capable of processing this data to ESPHome sensors, isn't it? One of the advantage I can think of myself, BLE monitor is supporting more sensors. than ESP Home. But are there other advantages?. 

I'll look into the  user permission check, thanks.

Magalex2x14:
ESPhome device receives BLE advertisement, and without further processing sends it via the local network to HomeAssistant, where ble_monitor parses it. Cool solution! For example, in this way, you cannot have a bluetooth interface at all on the HA host, but catch the data with ESP32 boards scattered around the rooms.

myhomeiot:
> Nice work, I will try to test is a.s.a.p. If I understand it correct, you use an ESPHome device to receive BLE messages and than broadcast the same data again? So it is working as some kind of "range extender".
> 
> Just curious, what is the benefit of this, ESPHome is also capable of processing this data to ESPHome sensors, isn't it? One of the advantage I can think of myself, BLE monitor is supporting more sensors. than ESP Home. But are there other advantages?.
> 
> I'll look into the user permission check, thanks.

It's not urgent at all, so take your time ;-) Most users which want to test it and use already did it manually with this PR.
I forget to tell you in previous message, please also change the service name, etc if needed, I will change everything accordingly.

Yes, I use ESPHome devices to receive BLE messages, add HCI [header](https://github.com/myhomeiot/esphome-components/blob/9db23eade2a6faf8a57b5171abdb3de42ddccf7a/components/ble_gateway/ble_gateway.cpp#L16-L34) and send it by WiFi to Home Assistant and your integration.

Yes, ESPHome can process BLE messages but has a limited devices support. Unload BLE processing to external component helps to free the ESP memory. BLE stack in ESPHome use relatively huge amount of RAM and sometimes if you have on top of BLE stack 10-15 BLE sensors components its start reboot spontaneously. BLE Gateway it's a one components with low code and memory footprint which process data from all sensors.

Sure in the case when ESP, for example, control the heater for reliability you should have temperature reading on board and this is not a solution. But in case when your heater connected to Home Assistant or in case when you just need extend BLE coverage it's a better solution.

But, for my opinion **the main pain** with regular ESPHome BLE sensors components (why I create BLE Gateway component) that if you have BLE device which moves from area to area you should have BLE sensor component for this device at the all ESPHome devices. **In this case in Home Assistant you will have all these sensors entities from all ESPHome devices** :-)
With BLE Gateway I can have one package yaml (like in Home Assistant) with BLE Gateway configuration and all my BLE devices macs, which I can include into all my ESPHome devices configurations and forget about this problem ;-)

myhomeiot:
> ESPhome device receives BLE advertisement, and without further processing sends it via the local network (WiFi) to HomeAssistant, where ble_monitor parses it. Cool solution! For example, in this way, you cannot have a bluetooth interface at all on the HA host, but catch the data with ESP32 boards scattered around the rooms.

Exactly ;-)

Ernst79:
I tried to add your suggestion about the user permission check, But I get the following error. Most likely, as we have defined in in `async_setup`, while your example has it in `async_setup_platform`. 

```
  File "/usr/src/homeassistant/homeassistant/setup.py", line 254, in _async_setup_component
    result = await task
  File "/config/custom_components/ble_monitor/__init__.py", line 191, in async_setup
    platform = entity_platform.async_get_current_platform()
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 740, in async_get_current_platform
    raise RuntimeError("Cannot get non-set current platform")
RuntimeError: Cannot get non-set current platform
```

Ernst79:
Tested the example, works perfect. device is added as expected

C:\Users\903494\Pictures\parse_data_service_screen.png

I do think that we need to add some documentation for this new feature. I'll merge this part and create a page in the docs. May I ask you to fill this page?

Ernst79:
I added the page here, 

https://custom-components.github.io/ble_monitor/parse_data

Could you check if you want some more info? 

I have an issue with the automation example, to put it in the docs, the jekyll docs are hiding everything between double `{{ xxx }}`, so the example isn't correct. Not sure how to fix this,

myhomeiot:
> I tried to add your suggestion about the user permission check, But I get the following error. Most likely, as we have defined in in `async_setup`, while your example has it in `async_setup_platform`.

I will try to implement this and will make PR.

> I do think that we need to add some documentation for this new feature. I'll merge this part and create a page in the docs. May I ask you to fill this page?

I can try, if you make small sketch with list of points which you think should be explained.

Ernst79:
I already made a first start of the page, I think the page should explain what the service does and how it can be used. I would only explain the BLE monitor side here, and provide some links to your ESPHome BLE gateway. 

You can just edit this page, 
https://custom-components.github.io/ble_monitor/parse_data

myhomeiot:
> I tried to add your suggestion about the user permission check, But I get the following error. Most likely, as we have defined in in `async_setup`, while your example has it in `async_setup_platform`.

Yes, it's look more challenged and requires lot of changes. If you like Nabucasa way and when you have the inspiration to do refactoring you can do it as they [suggest](https://github.com/home-assistant/core/blob/dev/homeassistant/components/alarm_control_panel/__init__.py#L58-L68).
Currently not many peoples aware and use Home Assistant user permissions, so it's definitely can wait ;-)

Thanks you for your work!

Ernst79:
6.2.0 has been released

myhomeiot:
> 6.2.0 has been released

Thanks you!
I added the way when ESPHome devices will get MAC addresses directly from BLE Monitor configuration, so it's will make the configuration even [simpler](https://github.com/myhomeiot/esphome-components#advanced-configuration-where-esphome-devices-gets-mac-addresses-from-passive-ble-monitor-configuration).

PR #573, opened 2021-12-04
Update parse_data.md

myhomeiot:
For now this documentation looks enough, I will update it accordingly users questions if they have any.

I fix automation example and now it's should looks ok.


Ernst79:
Thanks, I’ll create a new release tomorrow

PR #597, opened 2021-12-30
Add support for b-parasite

jhbruhn:
See here https://github.com/rbaron/b-parasite

Open Source DIY Plant sensor using BLE. It already has an ESPHome integration but this seems to be cleaner.

It uses the same uuid16 as the Xiaomi ATC sensors. Currently im identifying those via the message length, but if you have a better idea please correct me there.


custom_components/ble_monitor/const.py
```diff
@@ -588,6 +588,7 @@ class BLEMonitorBinarySensorEntityDescription(
 # - [instantly updating sensor list]:   sensors that update the state instantly after new data
 # - [binary sensor list]:               binary sensors
 MEASUREMENT_DICT = {
+    'b-parasite'              : [["temperature", "humidity", "moisture", "voltage", "rssi", "illuminance"], [], []],
```
Ernst79, line 591: Move this line to the same position in the dictionary as you do in the next dictionary, where you have put this just below the Tempo Disc THPD sensor

Do not use the "brand", but use a name of the specific sensor, e.g. `Plant sensor v1`, as I wrote in an earlier comment.

custom_components/ble_monitor/const.py
```diff
@@ -748,6 +749,7 @@ class BLEMonitorBinarySensorEntityDescription(
     'Moat S2'                 : 'Moat',
     'Tempo Disc THD'          : 'BlueMaestro',
     'Tempo Disc THPD'         : 'BlueMaestro',
+    'b-parasite'              : 'rbaron',
```
Ernst79, line 752: This will become, is you follow my previous comments

```
'Plant sensor v1'              : 'B-Parasite'
```

custom_components/ble_monitor/const.py
```diff
@@ -758,6 +760,7 @@ class BLEMonitorBinarySensorEntityDescription(
 # Selection list for report_uknown
 REPORT_UNKNOWN_LIST = [
     "ATC",
+    "b-parasite",
```
Ernst79, line 763: This should be the "brand", so in your case `rbaron` or `Rbaron`

custom_components/ble_monitor/ble_parser/bparasite.py
```diff
@@ -0,0 +1,87 @@
+"""Parser for BParasite BLE advertisements"""
+import logging
+from struct import unpack
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def parse_bparasite(self, data, source_mac, rssi):
+    """Check for adstruc length"""
+    device_type = "b-parasite"
```
Ernst79, line 10: You will need to add the device type here. The device type can't be the same as the "brand". E.g. for Xiaomi sensors, the device type is e.g. called LYWSD03MMC, not Xiaomi. My proposal

Manufacturer = `B-Parasite`
Device Type = `Plant sensor v1` (up to you to decide, but keep in mind that you might want to create other sensors in the future, so a version number might be useful)
Firmware = `B-Parasite v1` or `B-Parasite v2`

custom_components/ble_monitor/ble_parser/__init__.py
```diff
@@ -84,8 +85,11 @@ def parse_data(self, data):
                     if uuid16 == 0xFFF9 or uuid16 == 0xFDCD:  # UUID16 = Cleargrass or Qingping
                         sensor_data = parse_qingping(self, adstruct, mac, rssi)
                         break
-                    elif uuid16 == 0x181A:  # UUID16 = ATC
-                        sensor_data = parse_atc(self, adstruct, mac, rssi)
+                    elif uuid16 == 0x181A:  # UUID16 = ATC or b-parasite
```
Ernst79, line 88: This is a design decision in your firmware, using the ATC UUID16. At the moment, this will work, but if the ATC firmware devs decide to add an extra byte in the future, it might bite with your sensors (ATC already has 4 formats with different length). You might want to decide to change the UUID to something else in the future. UUID are normally reserved for actual companies, most "unofficial" BLE implementations use the format `FF` + 2 bytes as a company identifier, see e.g. the Govee sensors. I do understand if you don't change this now, but you might want to consider this for future development.

Ernst79 (changes requested):
Thanks, I have reviewed your PR. Please have a look, especially about the decision what the "brand" is and what should be the name of your "device"

custom_components/ble_monitor/ble_parser/__init__.py
jhbruhn: 0x181A is not allocated to ATC or any company. It rather is a generic UUID16 for a "Environmental Sensing" GATT Service. Thus any device may publish using that UUID, not only ATC (https://btprodspecificationrefs.blob.core.windows.net/assigned-values/16-bit%20UUID%20Numbers%20Document.pdf). Maybe there is some other identifier (besides mac address) usable here?

Ernst79: Yes, I know, I just want to warn you that you might get into trouble if the ATC devs decide to make a 20 byte long advertisement. But it's ok to leave it like it is for now, and change it if the ATC devs decide to change the format in the future. I have no idea if they have any plans to do. 

Alternative is, in stead of `16181A` is using `FFxxyy`, where for `xxyy`, you choose something that is not in this list 

https://www.bluetooth.com/specifications/assigned-numbers/company-identifiers/

custom_components/ble_monitor/ble_parser/bparasite.py
jhbruhn: The manufacturer in this case would be rbaron, the creator of the device. There are multiple versions of the same device (b-parasite), with and without illuminance (-> Device Types). They are called b-parasite V1.0.0 and b-parasite V1.0.0. There is a third revision, V1.2.0, but that had only hardware changes and does not affect the communication. I have changed this accordingly.

custom_components/ble_monitor/ble_parser/__init__.py
jhbruhn: I understand, sorry for sounding condescending. I am not the creator of the b-parasite device, but changing the UUID would break a lot of already deployed devices. This seems weird from a BLE-spec perspective, but maybe I'm missing something. 
If this breaks in the future, perhaps we can make use of the first two bytes of the data to differentiate between ATC and b-parasite (https://github.com/rbaron/b-parasite/tree/main/code/b-parasite#bluetooth-low-energy-advertisement-data-encoding)

custom_components/ble_monitor/ble_parser/bparasite.py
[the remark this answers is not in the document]
jhbruhn: I don't think there is a specific "brand"-identity here, but in its documentation, the name is always written in lowercase letters, so i'd keep it that way.

custom_components/ble_monitor/ble_parser/__init__.py
Ernst79: ah, I though you were the creator of these b-parasite device. As said before, no problem to leave it like it is. If ATC format changes, we can look into it how to solve.

custom_components/ble_monitor/ble_parser/bparasite.py
Ernst79: Ok, your change makes sense, in that case.

Ernst79 (approved):

Ernst79:
Added in 6.4.2-beta.

jhbruhn:
Perfect, thanks for maintaining!

PR #599, opened 2022-01-03
Fix instance member access

daniele-athome:
I'll go out on a limb here and guess the right instance member :-)


Ernst79:
Thanks!

PR #602, opened 2022-01-04
Fix type of const.PROBES

nattofriends:
Presumably this is meant to be a tuple, not a single string


Ernst79:
Thanks, will be included in the next release

PR #635, opened 2022-01-18
Jinou BEC07-5 beacon  temperature & humidity sensor

marcomiani:
Hi
I have a Jinou BEC07-5 beacon.  It came with the following 'manual'

`Temperature and humidity sensor service: UUID 0XAA20 
There are three characteristics:
Service 0XAA20. Temp & humid data. There are 6 bytes.
1. Temperature positive/negative: 0 means positive (+) and 1 means negative 
(-)
2. Integer part of temperature. Show in Hexadecimal.
3. Decimal part of temperature. Show in Hexadecimal.
4. Reserved byte. Ignore it.
5. Integer part of humidity. Show in Hexadecimal.
6. Decimal part of humidity. Show in Hexadecimal.
For example: 00 14 05 22 32 08 means +20.5C 50.8%
01 08 09 00 14 05 means -8.9C 20.5%
0XAA22(Read,Write)：start/stop checking temp and humid. 0x01(start), 
0x00(stop).
0XAA23(Read,Write)：checking frequency, unit(10ms), default is 100, that is 
checking every second.
The service UUID in broadcasting can be used as filter condition when searching`

I have it working in my HA installation , temp & hum values verified with the android app shipped with the beacon

The value of temperature and humidity is represented with two bytes, one for the integer part and one for the decimal part.
To combine these values I concatenated them as strings and then converted to double, an alterative is to transform the decimal part  dividing by 10 if the value is <=10  or dividing by 100 if the value is >10 and then add the integer part.

I hope to have implemented it in a way you approve.

thank you


Ernst79:
Thanks, I will look into it a.s.a.p.

Ernst79 (approved):

PR #646, opened 2022-01-22
Add support Beacon UUID and iBeacon

dmamontov:
Quite a big improvement, adds the ability to track the indication of the Beacon UUID as a separate identifier, instead of MAC. The matter is that MAC in such devices can be dynamic.

Also added support for the iBeacon protocol.
Many more may be added in the future.

Below is an example of an android phone in which the home assistant emulates the iBeacon protocol:
image


custom_components/ble_monitor/__init__.py
```diff
@@ -513,9 +522,9 @@ def __init__(self, config, dataqueue):
         if self.config[CONF_DEVICES]:
             for device in self.config[CONF_DEVICES]:
                 if CONF_DEVICE_ENCRYPTION_KEY in device and device[CONF_DEVICE_ENCRYPTION_KEY]:
-                    p_mac = bytes.fromhex(device["mac"].replace(":", "").lower())
+                    p_id = bytes.fromhex(dict_get_or_clean(device).lower())
```
Ernst79, line 525: Just that I understand it, can `p_id` either be a MAC address or an UUID?

dmamontov: everything is right. there are types of beacon with encryption. I want to add more of them, but for now I added ibeacon as an example

Ernst79:
It looks good to me, I propose we make a beta release first, to test it.

Ernst79:
Released as 7.3.0-beta

dmamontov:
Fine. Thanks.

PR #648, opened 2022-01-23
Added sv.json

el97:
Added a Swedish translation


Ernst79:
Thank, will include it in the next release

PR #654, opened 2022-01-23
added support altbeacon

dmamontov:
* Added AltBeacon support. Added support for manufacturer indication in sensors. for example with AltBeacon, the manufacturer is specified in an additional 2 bytes.

* Minor iBeacon refactoring.

image


Ernst79:
Excellent work! I propose to keep it in beta a few days. if you want to add more protocols, we can just add them, and make a final release by the end of this week. 

I made a few small changes in #653, which just got merged before this one.

dmamontov:
While there are ideas for other protocols and devices, for example, Google decided to get rid of everyone), but we need to think about the implementation. How can you and merge)

In general, I think in the component you can implement not mac and uuid, but filters, and you can add different ones, mac, uuid, namespace, name, etc. but this is a big improvement that will break backward compatibility. In principle, I can implement, but I need to think it over and agree with you)

Ernst79:
Sure. Note that I was approached by one of the devs of HA that they wanted to make BLE monitor an official integration, but they wanted to do some rewriting. BLE monitor is the result of years of extending the original functionality, so it might not be as logic as it should be. Moreover, I'm not an experienced programmer, actually, I'm just an civil engineer who is doing this as a hobby. So my coding is probably not as it should be 😄 

I haven't heard back from the HA dev what the plans are, and on what term, but keep this in mind if you decide to a lot of work.

dmamontov:
Thank you. I'll keep it on mind)

custom_components/ble_monitor/ble_parser/__init__.py
```diff
@@ -146,6 +147,10 @@ def parse_data(self, data):
                     if comp_id == 0x4C and man_spec_data[4] == 0x02:  # iBeacon
                         sensor_data, tracker_data = parse_ibeacon(self, man_spec_data, mac, rssi)
                         break
+                    elif man_spec_data[0] == 0x1B and man_spec_data[1] == 0xFF \
```
Ernst79, line 150: I don't think you need `man_spec_data[1] == 0xFF` in this line, as `0xFF` in this position is by definition a packet with `manufacturer specific data`

dmamontov: yes, the tests pass, there should be no problems.

dmamontov: https://github.com/custom-components/ble_monitor/pull/655

PR #663, opened 2022-01-26
added portuguese brasilian language support

hudsonbrendon:


Ernst79:
Thanks, I will include it in the next release

PR #703, opened 2022-02-05
New bt helpers

Magalex2x14:
I checked it on rpi3b HomeassistantOS and on HA core with three interfaces - it looks good.

The only problem I see so far is:
```
2022-02-05 19:31:35 ERROR (MainThread) [custom_components.ble_monitor] Bluetooth adapter with MAC address ['B8:27:EB:77:3C:EF'] was not found. It is therefore changed back to the default adapter. Check the BLE monitor settings, if needed.
2022-02-05 19:31:35 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Bluetooth Low Energy Monitor for ble_monitor
  File "/config/custom_components/ble_monitor/__init__.py", line 308, in async_setup_entry
    list(BT_INTERFACES.values()).index(bt_mac)
ValueError: 'B8:27:EB:77:3C:EF' is not in list

During handling of the above exception, another exception occurred:

  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 327, in async_setup
    result = await component.async_setup_entry(hass, self)
  File "/config/custom_components/ble_monitor/__init__.py", line 320, in async_setup_entry
    list(BT_INTERFACES.values()).index(DEFAULT_BT_INTERFACE)
ValueError: 'disable' is not in list
```
The current release has the same problem.
Reproduced on this branch with an empty dictionary before if BT_INTERFACES in bt_helpers.py:
image

Or a forced exception after hci_get_mac in the current release:
```python
raise IndexError("Test!")
```

Thus, the situation of interface loss is simulated. As far as I understand, just go to the integration settings and enable the Don't use Bluetooth adapter option. It seems to me that this can be done automatically, but I cannot navigate the config_flow code - I think you orient yourself faster.

In general, I think the code is ready for beta.


Magalex2x14:
Hmm...
The test fails with `Unable to open PF_BLUETOOTH socket` error :) How to ignore this?

Magalex2x14:
I solved the problem with an error when the interface is lost.
When connecting the adapter back without restarting the system, the adapter index is not restored - hci3 appears when hci0 is missing (and for it to appear in the ble_monitor list, you need to restart HA).
Restarting the system restores the indexes, additional manipulations with ble_monitor are not required.

Magalex2x14:
For a failed test we need to contact @ludeeus?

Ernst79:
Maybe (just maybe) it helps if you add the requirements in master. Not sure if it uses the requirements from master. Havent looked it it though

Magalex2x14:
I think it will still fail.
As far as I understand, the essence of the problem is that the tester cannot execute `BT_INTERFACES = hci_get_mac([0, 1, 2, 3])` in bt_helpers.py, because the test system does not have sockets of type `PF_BLUETOOTH`.
I think we need to tinker with the `.github/workflows/test_commit.yml` file

Magalex2x14:
To specify somehow to ignore the execution of hci_get_mac function

Ernst79:
I see, it has installed the requirements as you specified. Dont know how to exclude a certain function from the test. If you cant find it, I can have a look tomorrow. Otherwise a try except could be a solution perhaps

Magalex2x14:
> Otherwise a try except could be a solution perhaps
Yes, I guessed) It is necessary to consider the test system as a user one )
Accidentally disturbed ludeeus in vain - was confused, stick the @ symbol 😅

Ernst79:
Will test tomorrow

Ernst79 (approved):
Just tested it, works fine. I'll create a beta release to test it for a while.

Ernst79:
Do you want me to wait with the beta release, or can it be released?

Magalex2x14:
I think its ready, release it

Issue #709, opened 2022-02-11
Bluetooth dongle not recognized with newest update


permissionBRICK:
I updated to 7.6.0, and was greeted with all BLE sensors no longer working, it turns out the Integration fails to discover my BLE stick.
![image](https://user-images.githubusercontent.com/40219477/153661769-4617a722-2818-4fb4-9469-6320f715136e.png)

```
2022-02-11 20:51:03 WARNING (MainThread) [custom_components.ble_monitor.bt_helpers] hci%i has no (or disabled) BT LE capabilities.
2022-02-11 20:51:04 ERROR (MainThread) [custom_components.ble_monitor] Bluetooth adapter with MAC address ['XX:XX:XX:XX:XX:XX'] was not found. It is therefore changed back to the default adapter. Check the BLE monitor settings, if needed.
2022-02-11 20:51:04 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Bluetooth Low Energy Monitor for ble_monitor
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/config_entries.py", line 327, in async_setup
result = await component.async_setup_entry(hass, self)
File "/config/custom_components/ble_monitor/__init__.py", line 371, in async_setup_entry
list(BT_INTERFACES.values()).index(DEFAULT_BT_INTERFACE)
ValueError: 'disable' is not in list
```
Restarting HA or even the Host PC does not help the situation. I am Running HA Docker on Unraid, and am running the Container in privileged mode to pass through the BLE stick.
lsusb does show the device on the host, as well as inside HA:
Bus 001 Device 002: ID 0a12:0001 Cambridge Silicon Radio, Ltd Bluetooth Dongle (HCI mode)

I downgraded to 7.5.2 and it is working again.

Magalex2x14:
Hello! Can you, please, install 7.6.0 and enable [debug logging](https://custom-components.github.io/ble_monitor/faq#debug), and throw here everything that appears in the log before the error since the start of HA.

It would also be useful to see the output of the `sudo btmgmt info` command on your host.

permissionBRICK:
Here you go

```
[...]

PR #711, opened 2022-02-12
correct check for adapter LE capability

Magalex2x14:
Inspired by #709
Had to make a new 7.6.1 urgent release with LE capability checking removed (as in the old bt_helpers) because the number of affected users could be large.

This pull request contains the correct LE capability check procedure and minor corrections to the log messages.
Tested with my old BT 2.0 adapter.

I am convinced that these changes can be safely included in the next release.


Ernst79 (approved):

Ernst79:
Tested, works fine. Thanks.

PR #725, opened 2022-02-18
Midea [WIP]

Ernst79:
## Add support for Midea sensors

- [x] Add filter for Midea sensors
- [ ] Add parser for Midea


Ernst79:
Closing this PR, as the device requires a connection to be set up, which is not supported by BLE monitor

PR #776, opened 2022-03-15
b-parasite: Fix parsing of negative temperatures, support protocol v2

jhbruhn:
This PR adds a fix for negative temperatures reported by b-parasites, which were treated as unsigned before, but are actually.
Also this adds support to protocol version 2 which increases the range of the temperature in exchange for (unnecessary) precision.

Another thought about the UUID used by both this device and the ATC firmware:
The packets from b-parasites include the MAC address of the device. Would it be a feasible solution to compare that MAC address (at a known location in the advertisement data) with the one provided from the bluetooth "receiver"? A known limitation is that macOS/iOS spoof the mac address which would lead to the packets not being decoded there, but otherwise it is the best solution for now, other than switching to a different UUID. Any thoughts?

### DIFF
--- a/custom_components/ble_monitor/ble_parser/bparasite.py
+++ b/custom_components/ble_monitor/ble_parser/bparasite.py
@@ -8,13 +8,13 @@ _LOGGER = logging.getLogger(__name__)
 def parse_bparasite(self, data, source_mac, rssi):
     """Check for adstruc length"""
     msg_length = len(data)
-    if msg_length == 22:  # TODO: Use version bits?
+    if msg_length == 22:  # TODO: Use protocol bits?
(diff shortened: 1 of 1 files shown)


Ernst79:
I would suggest to leave it as it is. There is no issue with the ATC firmware at the moment due to the different length. Another possibility could be to add an AD type 'complete local name' (`0x09xxxxxx`). The bytes can be an abbreviation of the name (e.g. bpa) (`04 09 62 70 61` translates to length 4, type 9 (= complete local name), 62 = b, 70 = p, 61 = a). I can than use this as an additional filter.

Ernst79:
Your fix is included in 8.0.4-beta

PR #807, opened 2022-03-30
Reverse xiaomi binary sensors lock states & decouple locks for ZNMS17LM.

vincent-k:
This fixes two main issues I faced with Xiaomi Aquara N200 doorlock (ZNMS17LM):
- lock states are reversed on HASS binary sensor: 0=>locked 1=>unlocked.
  Note: this commit applies the fix for all Xiaomi locks and must be therefore double checked.
- having lock, anti-lock & child-lock events reporting on same sensor leads
  to incorrect global state. Eg. when the door is fully locked (main lock + anti-lock)
  if we release only the anti-lock then the sensor goes to unlocked state which is
  incorrect. This commit decouples the lock into three distinct sensors: 'lock',
  'anti-lock' & 'child-lock' to avoid such issues.
  Note: this only targets 'ZNMS17LM' model to keep a selective approach (some users may
  prefer the global version and some other Xiaomi locks may not be reporting those states).


Ernst79:
Thanks, I’ll look into it later.

Ernst79:
Could you check if the Add door sensor for ZNMS17LM also sends a "door" event, see #811 and #812. I guess the ZNMS17LM also sends that event.

Ernst79:
I have released 8.5.0-beta with this PR, for testing. I don't have a lock myself, so I can't test it myself. Please also check my previous question, we might be able to add a "door" sensor as well, if the ZNMS17LM is also reporting this (like the ZNMS16LM). This has not been added in the beta, as I first need some confirmation that this door events are available on the ZNMS17LM.

vincent-k:
Hi,
Great! I confirm that door events are also reported by ZNMS17LM:
Screenshot_20220405-231756_Home Assistant

Ernst79:
I'm closing your PR, as it was included in #812. So, it will be added in the next release.

PR #842, opened 2022-04-23
Bug fixes: Oral-B (IO Series 7)

codyc1515:
This change introduces a new way to map the "Modes" values based on the device model.

Support for the SmartSeries 7000 has been retained and support for the IO Series 7 has been added. It is likely that some mappings are missing for IO Series 8 & 9, however I do not have these devices on-hand.

The value for the model should really go into the dveice_type field, however this caused the whole plug-in to break for me (attributes stopped reporting). I instead placed the model into the "Firmware" field which worked and would've otherwise just been "Oral-B". The second issue with this is that we do not know the model until a broadcast is received.

Open to suggestions on ways to improve this commit as I am not a Home Assistant, BLE or Python expert by any stretch.

### DIFF
--- a/custom_components/ble_monitor/ble_parser/oral_b.py
+++ b/custom_components/ble_monitor/ble_parser/oral_b.py
@@ -12,6 +12,7 @@ STATES = {
     4: "charging",
     5: "setup",
     6: "flight menu",
+    8: "selection menu",
     113: "final test",
     114: "pcb test",
     115: "sleeping",
--- a/custom_components/ble_monitor/ble_parser/oral_b.py
+++ b/custom_components/ble_monitor/ble_parser/oral_b.py
@@ -19,18 +19,6 @@ STATES = {
     116: "transport"
 }
 
-MODES = {
-    0: "off",
-    1: "daily clean",
-    2: "sensitive",
-    3: "massage",
-    4: "whitening",
-    5: "deep clean",
-    6: "tongue cleaning",
-    7: "turbo",
-    255: "unknown"
-}
-
 PRESSURE = {
     114: "normal",
     118: "button pressed",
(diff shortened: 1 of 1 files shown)


AgenttiX:
Thank you for this pull request! I've been wondering for quite a while whether the iO Series has been or could be supported. Based on the [code in your pull request](https://github.com/custom-components/ble_monitor/pull/842/commits/75b02894c34eee3e64a5b0d5377edc1cef454084) it seems that the modes are different in the iO Series compared to the SmartSeries 7000. Is there also a difference in the pressure data, or have you added the [pressure mappings](https://github.com/custom-components/ble_monitor/pull/842/commits/96d89de361514a25352b6e2df9d186d9f7ce65b1) just for clarification? Does this pull request decode all the data sent by the iO Series 7, or are there still data fields that are not yet supported?

It seems that the documentation for the state of Oral-B toothbrush support is missing at the moment, as described in [this issue](https://github.com/custom-components/ble_monitor/issues/844).

codyc1515:
Hi.

Yes - the modes are different and overlap between the two. I am not sure why Oral-B made this decision but it is the reason that I have split the two out. For me I was seeing the wrong or unknown values populating the attributes in HA, so I have simply obtained the mapping information by pressing every button and watching the values outputted over BLE.

The pressure values were not mapped before and were simply a number. For me these values correspond in the app to normal (green colour) and high (red colour), however the value for pressure also changes when you press the button strangely enough. I have added that too.

I believe that all of the passive data from the BLE advertisement should be mapped for the iO Series 7, however it is likely that the Series 8 and so on have different mappings too as they have a different feature set available.

Thank you

Ernst79:
Thanks, I will review it in a couple of days. First have to unpack my stuff in my new house.

Ernst79:
You will have to use the device type to define the sensor type (SmartSeries 7000, etc), firmware should be Oral-B. Next, you have to add the same device types in the const.py file, where you can define all sensors for a specific device. And there is a second dictionary that couple the device type to the manufacturer. 

If you can’t find it, I will change it for you in a couple of days, no problem

codyc1515:
Sorry, this was the extent of my abilities so I'm not sure what you're referring to.

Agree the model should be in device type but would appreciate your help with doing that.

Good luck with the move.

Ernst79:
I have created a new PR, as it was easier to work on for me. I copied your changes into #847 and flipped the device_type and firmware. I will close this PR and merge the other one. Will release it afterwards.

PR #892, opened 2022-06-11
Update Telink Flasher url

grilix:
I've had quite some issues using the page linked here. However, I've found this other one on the same repo worked perfectly: https://github.com/atc1441/atc1441.github.io/blob/master/TelinkFlasher_all_in_one.html


Ernst79:
Thanks!

PR #916, opened 2022-07-04
Added new lock support: lockin Q2

neroxps:
This fingerprint lock is the Xiaomi Bluetooth protocol, but the event definition is not completely parsed according to the Xiaomi door lock definition.
The event of locking the door outside the door is an arming event using the Mijia Bluetooth protocol.


custom_components/ble_monitor/ble_parser/xiaomi.py
```diff
@@ -247,6 +269,22 @@ def obj000b(xobj, device_type):
...
+                "action": action,
+                "method": method,
+                "error": error,
+                "key id": key_id,
+                "timestamp": timestamp,
+            }
+            if method == "password":
+                if 5000 <= key_id < 6000:
+                    returnDict.update({"method": "one-time password"})
+            return returnDict
```
Ernst79, line 286: Isn't it much shorter by just overwriting `method`, like this. 

```python
        #Biometric unlock then disarm
        if device_type == "DSL-C08":
            if method == "password":
                if 5000 <= key_id < 6000:
                    method = "one-time password"
```

custom_components/ble_monitor/ble_parser/xiaomi.py
```diff
@@ -204,6 +205,27 @@ def obj0007(xobj):
...
+        return{
+            "lock": value,
+            "locktype": 'lock',
+            "action": 'lock outside the door',
+            "method": "manual",
+            "error": None,
+            "key id": None,
+            "timestamp": None,
+        }
+    return returnData
```
Ernst79, line 228: If I understand this correct, your lock ("DSL-C08") will always return the return function, never the `returnData`. Why did you than implement the `armed away` and `timestamp`? Is that a possible outcome for other sensors that aren't implemented yet (if yes, than it is OK, just curious).

custom_components/ble_monitor/const.py
```diff
@@ -319,6 +319,15 @@ class BLEMonitorBinarySensorEntityDescription(
...
     ),
+    BLEMonitorBinarySensorEntityDescription(
+        key="armed away",
+        sensor_class="BaseBinarySensor",
+        update_behavior="Instantly",
+        name="ble armed away",
+        unique_id="arm_",
+        device_class=BinarySensorDeviceClass.LOCK,
+        force_update=True,
+    ),
```
Ernst79, line 330: See previous comment, this armed away sensor isn't used currently. It's fine to add it like you did, just curious if this is as you intended.

Ernst79:
Thanks @neroxps Looks great!

The only thing missing is a test. Just add a test here for your sensor

https://github.com/custom-components/ble_monitor/blob/master/custom_components/ble_monitor/test/test_xiaomi_parser.py

custom_components/ble_monitor/ble_parser/xiaomi.py
neroxps: According to the Mi Home protocol, the disarm event is timestamped, but the disarm event of my lock is timestamped without.

custom_components/ble_monitor/const.py
neroxps: Because my lock is using the arming event to pass outside the door to lock the door (handle to lift)
I was previously using this sensor to get the lock status, but have since eliminated this practice.

neroxps:
> 

Do I need to provide `aeskey` for the test code? I don't want to publish my `aeskey`.

custom_components/ble_monitor/const.py
Ernst79: Ok, but should we remove it than, to keep things clean. ?

Ernst79:
Ah, yes, it is normally needed to make a test run, but I can understand you don't want to publish a key from a lock 😄 We will do it without a test than, no problem.

Ernst79:
Just one question left, should we removed the "armed away" sensor definition in const.py, as we don't use it anyway. Or is there a reason to keep it?

neroxps:
> 

Okay, I'll remove the "armed away" sensor

Ernst79:
Thanks!

Ernst79:
Released as 8.9.4.

PR #920, opened 2022-07-05
Add dependabot

wrt54g:
I added dependabot for actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: weekly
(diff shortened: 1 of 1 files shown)


Ernst79:
Thanks

wrt54g:
@Ernst79 Now look at your PRs, because the bot has started already.

Ernst79:
Yes, I noticed that, have merged them. 

I’m new to dependable, I was expecting that it would e.g. update pycryptodomex in manifest.json, which has a newer version 3.15.0. But it doesn’t show that.

Ernst79:
Ah, ok. Will look into that later. I don’t want to update pycryptodomex anyway, as there is some bug that it isn’t working in Home Assistant.

Issue #953, opened 2022-07-17
New device_id for Switchbot Meter Plus


gilcu3:
### What happened?

My new switchbot meter plus sensor began providing `device_id = 0xc069` instead of `0x0069`. If this value is added to the list of accepted values, everything works correctly.

### Sensor type

Switchbot

Ernst79:
Fixed in 9.0.2. thanks for reporting

DigiH:
Hi @Ernst79, sorry to peak in again :blush: but doesn't including the **Status** byte into the **device_id** not possibly eventually  just require including all possible variations **00-FF**?

Admittedly, **Bit[7:4] - Reserved** might not use all bits, but even for a user adding the Meter (Plus) to one or more groups **Bit[3:0]** would mess with the **device_id**.

https://github.com/OpenWonderLabs/SwitchBotAPI-BLE/blob/latest/devicetypes/meter.md

Hope I'm not overstepping.

Ernst79:
I see, thanks. I was not aware of this. I guess we can define the device based on the second byte only. I’ll reopen this issue and will fix it.

DigiH:
I even spotted a mistake with the temperature decimal not ignoring the alters bits in our decoder when going through this again.

Win win for both of us ;)

Ernst79:
9.0.7 has been released, thanks for the input

DigiH:
Sorry @Ernst79 , I really don't want to be a pain, but I think your temperature could do with a little tweak as well.

In your syntax probably masking & 127 instead of the -128, and if conditioning when bit 7 is 0 = zub zero  temperatures * -1, otherwise as is, hence your current -128 only allowing for +0ºC temps and having to subtract 128 because bit 7 is 1.

And also the same problem I discovered above, with your byte2 also being messed up by bits 4-7, as soon a user has set alerts, encoded in those bits.

Unless I'm totally misreading your decoding.

I've tried to include all those eventualities in our test cases

https://github.com/theengs/decoder/blob/development/tests/BLE/test_ble.cpp#L125

https://github.com/theengs/decoder/blob/development/tests/BLE/test_ble.cpp#L364

Ernst79:
Any improvements are always welcome. I’ll fix that later, will probably not be the coming weeks, due to other things I have to do. Will reopen the issue to not forget

Ernst79:
@DigiH Do you know how to check the 7th bit? Is this correct?

```python
[...]

PR #954, opened 2022-07-19
Add new device type for switchbot TH plus

Ernst79:
Add new device type for switchbot TH plus, fix for #953


gilcu3:
Hi, on line `19` you also needed to add the id of the device :)

Issue #952, opened 2022-07-17
[New Sensor]: SmartDry dryer senror


yoavf:
### Sensor Description

SmartDry (RIP) clothes dryer BLE sensor (only one type)

### Additional information

The [SmartDry](https://www.connectedlifelabs.com/meetsmartdry) product cloud service [went dead](https://home-assistant-guide.com/2022/06/23/smartdry-goes-bust-and-creates-a-bunch-of-e-waste/).

It uses a BLE sensor, and the messages have been reverse-engineered [here](https://community.home-assistant.io/t/clothes-dryer-automations/149017/123):

> The first 4 bytes is the “Temperature” in Celsius as a float
> The next 4 bytes is the “Humidity” percentage as a float
> The “Wake” value has only output 0 when the dryer is off or 7 when it is running.

And from the same message, here's the esp32_ble_tracker lambda function  
```
            id(raw_sensor).publish_state(format_hex(x));
            uint32_t temp = (x[0] + (x[1] << 8) + (x[2] << 16) + (x[3] << 24));
            uint32_t hum = (x[4] + (x[5] << 8) + (x[6] << 16) + (x[7] << 24));
            uint16_t shake = x[8] + (x[9] << 8);
            uint8_t batt = x[10];
            uint8_t wake = x[11];
            id(temp_sensor).publish_state((*(float *) &temp));
            id(hum_sensor).publish_state((*(float *) &hum));
            id(shake_sensor).publish_state(shake);
            id(batt_sensor).publish_state(batt);
            id(wake_sensor).publish_state(wake);
```

### BLE advertisements

```shell
2022-07-17 14:12:57 INFO (Thread-3) [custom_components.ble_monitor.ble_parser] BLE advertisement received from MAC/UUID 1a0500000a29: service data: []manufacturer specific data: [b'\x0f\xff\xae\x01\x9d\xe2\xdbAp#`B\x00\x00.\x07']local name: N6BPXUUUID16: None,UUID128: None
2022-07-17 14:12:59 INFO (Thread-3) [custom_components.ble_monitor.ble_parser] BLE advertisement received from MAC/UUID 1a0500000a29: service data: []manufacturer specific data: [b'\x0f\xff\xae\x01\xaf\xb6\xdbA\x103`B\x00\x00.\x07']local name: N6BPXUUUID16: None,UUID128: None
[...]

PR #955, opened 2022-07-19
Add first draft for SmartDry

Ernst79:
Add support for SmartDry cloth drying sensor #952 

**To Do**
- [x] Figure out how to parse the data to a float in stead of a large UINT
- [x] Add documentation
- [x] Add tests


DigiH:
Hi again @Ernst79 :) I see we're both working on the same SmartDryer decoder.

Here are my findings so far with only having the same information as you in 

https://github.com/custom-components/ble_monitor/issues/952

• The battery info there is wrong, in my eyes. I suppose there it expects battery **percentage**, but with the few actual SmartDry app user data I have the battery is reported in **voltage** and therefore needs to include the following byte as well for little endian decoding. Then have two calculations for getting the same voltage with the following wake  being true or false
• This extra following byte also indicates the wake state, with **00** being false/off and **07** being true/on, so this is how I decided to decode it, with wake true/false status.
• The shake seems fine
• Now to the temperatures and humidity, yes, still very intriguing. The suggested little endian for both doesn't make sense to me with the results for the few user data, but for humidity I sort of got some reasonable results with big endian and the subtracting from 100 to get the final precentage. Temperature still remains a mystery to my right now.

Not sure if you have already had a peek at my test decoder

https://github.com/DigiH/decoder/tree/smartdry-test

but might be an idea to collaborate on this and share our findings.

Ernst79:
Hi, yes, sure we can exchange experiences, but I have to say I don’t have the device myself, i just tried to solve one of the request for support for this device. So, I need as much help as you I guess. Let try to figure out how to decode this thing….

I will have a look at what you have found so far.

DigiH:
Yes, not having the device makes it difficult ;) here is also some comparison data which I produced with 

**Grey** - test cases with the available data from the HA forum discussion user
**Brown decoder results with green temp/hum** - decoder with big endian
**White** - the user's relevant cloud data (very likely with a big(ish) time gap to each raw data above)
**Brown decoder results with red temp/hum** - decoder with little endian as suggested in the original decoding

SmartDryData

So I'm very confident about my battery voltage finding, but the rest is very illusive, especially since the cloud data seems to have a substantial time lag to the original nRFConnect data readings from the user, apart from the very slow changing battery voltage ;)

@yoavf, you actually seem to own this device, is this correct?

If so, my suggestion would be to take it out of the dryer, let it settle in the room for a while to acclimatise. Then take readings where the temperature and humidity should be constant over time, and the shake value 0. Comparing this then to your raw data should make it a lot easier to confidently re-engieer, instead of chasing constantly changing values from within a dryer.

If the device stops sending raw data when **wake** is **00** for longer, a quick shake should wake it and still have the constant temp and hum ;)

Really would be great to get his decoded together, since they shut down their cloud service soon.

Ernst79:
I updated my code to be the same as yours now. 

@yoavf I can make a beta version of BLE monitor, if you want, such that you can compare the results of the readings of the app with the results in BLE monitor. In that way, it is may be easier to compare and find a relation between the values.

Ernst79:
Try with 9.0.6-beta and compare the results from the app with ble monitor results. 

Beta versions can be installed with HACS, but you must switch on the toggle to install beta-versions

Ernst79:
humidity seems to be correct, except in the 4th measurement. Might be just reported wrong/not at the same time. 

Temperature is indeed not logic. Hopefully the beta version allows @yoavf to make a comparison realtime

Ernst79:
Posted a question on the HA forum topic as well.

DigiH:
Just read your forum comment, thanks.

I'm still not very confident with the current temperature either, especially since I also had to change the resulting variable into a long long from long. Somehow doesn't seem realistic.

Looking forward to what answers we get from the forum and from @yoavf's testing.

Ernst79:
According to the forum response, it should be as follows

the temperature is 9bc8af41, so you have to take those 4 values in decimal and add them together to get 155 + 200 + 175 + 65 = 595, so you have 59.5 degrees Celsius which is 139.1 degrees Fahrenheit. 

Will try tomorrow.

DigiH:
Thanks :thumbsup: I already sort of tried that, but with their indicated left shifting, which didn't produce any viable results.

Will also test this without shifting, though still not sure how get this result with their shifting instructions.

Ernst79:
Yes, I agree I’m not convinced that this is correct. This left shifting normally is the same as little endian, I tried that also, same result.

DigiH:
Really best if yoavf, or someone else with the device, can do some more testing and verification, especially in a steady room climate environment with little to no variations in the temperature and humidity. Then we can confidently reverse engineer the values.

DigiH:
Just quickly added them up manually and added them to my comparison chart in orange.

Screenshot 2022-07-23 at 23 29 12

I don't think I'll even adjust the decoder, as it doesn't really look right to me. And with all the decoders I've done so far I've never once seen this adding 4 converted bytes together, while it could still be possible. I still think it might just be a usual 2 byte little endian temperature and humidity encoding somewhere, possibly even with the additional averagetemp and averagehum, but our best bet might be to wait for some device owner to help.

yoavf:
Thanks, I'll try running the beta today and report the values

yoavf:
OK, here's some data I grabbed by (manually) logging the data from the "Sensor Test" screen in the SmartDry app, matched with the beta `ble_monitor` output in the logs (hopefully at the same exact time, though there could be some minor differences):

[smartdry.csv](https://github.com/custom-components/ble_monitor/files/9180418/smartdry.csv)

Note that the app reports the temperature in F on this screen, so I used a formula to include the C value.

DigiH:
Thanks a lot @yoavf, while I am not as familiar with the ble-monitor log output the manufacturerdata somehow looks shorter to me. Might be due to some beta version setting?!? I'm sure sure @Ernst79 will know more.

> Note that the app reports the temperature in F on this screen, so I used a formula to include the C value.

Probably just an app setting, but also something I haven't considered at all before, that this sensor might actually report the temp in ºF, unlike all other temp sensors I have across so far. Unlikely, but with all the mystery this SmartDry has given us so far …

Ernst79:
it isn't shorter, the part that is missing is the local name. the service_data is just the 0x16 part. The full raw BLE data of the 7th reading in the excel table is `043E3902010000AD114BDC1B002d02010607093241485A44560fffae0124c2eb41000c60420200c406020A031107FB349B5F8000008000100000AA3A`

Temperature = 29.4C, 85 F
With a 4 byte UINT, little endian, temperature results in 1105969700
With 4 bytes added up, temperature results in 530
With 4 bytes UINT, big endian, temperature results in 616753985

mysterious, but I still don't see the relation....

DigiH:
Thanks for the clarification.

Still no obvious relationship I can see either.

I really hope that @1technophile asking the manufacturer for their encoding scheme has a positive reply. It should be in their interest to be able to give their existing customers some way for them to keep using their devices after their shut down.

Ernst79:
The 6 bytestrings of the BLe advertisements in the Excel
```
data_string = "043e2802010401f269352207ce1c11071bc5d5a50200b89fe6114d22000da2cb0916000d54006400990ec7"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae01203fee4150d55f420100c706020A031107FB349B5F8000008000100000AA3A"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae0144e7ed41008f5f420100c706020A031107FB349B5F8000008000100000AA3A"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae017a63ed4140315f420100c706020A031107FB349B5F8000008000100000AA3A"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae01ca9dec4160fc5f420100c306020A031107FB349B5F8000008000100000AA3A"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae0124c2eb41000c60420200c406020A031107FB349B5F8000008000100000AA3A"
data_string = "043E3902010000AD114BDC1B002d02010607093241485A44560fffae017de6ea41702360420000c406020A031107FB349B5F8000008000100000AA3A"
```

DigiH:
So the wake also seems to be **06** instead of the reported **00** in the forum discussion - if I remember correctly - when off.

All voltages between 1.732 - 1.735 V still seems reasonable.

The 3rd and 4th byte, after the company id, big endian, divided by 2000 gives a pretty close constant temperature in ºC, especially since it is calculated in the table.

ee41
ed41
ed41
ec41
eb41
eb41
ea41 = 59969 / 2000 = 29,9845

Magalex2x14:
Hello, guys! This data looks like a little endian float numbers
image

https://www.scadacore.com/tools/programming-calculators/online-hex-converter/

This roughly matches the data from the csv file above.

DigiH:
Ouch! we really missed that one with the INT little endian :( 

mea culpa! ;)

DigiH:
And both temp and humidity work fine :) so the forum definition was correct, even though the explanation with the addition without the little endian left shifting wasn't quite.

I think we should also tell them about their current battery percentage vs. the actual voltage decoding.

jmtorres:
> And both temp and humidity work fine :) so the forum definition was correct, even though the explanation with the addition without the little endian left shifting wasn't quite.
> 
> 
> 
> I think we should also tell them about their current battery percentage vs. the actual voltage decoding.

Hi. I was the person on the forum that posted that information. I was aware that it was in little endian format and actually spent too much time trying to flip the endianness before I realized that the esp was already handling the input as little endian and I just need to cast as a float. Based on reading the type code in esp, it would convert it automatically on devices that run with big endian, but I have not tested that theory.

I couldn't find what you were referring to with the battery voltage. What did you discover with that?

Let me know if I can be of any other help. I've spent quite a bit of time looking at this sensor.

DigiH:
> I couldn't find what you were referring to with the battery voltage. What did you discover with that?

Hi @jmtorres, funnily enough I was just pondering over the battery voltage again.

In the forum it was posted that only the one byte should be the battery percentage, but even with some of the examples, I think it was 68, it would have been over 104%.

The user Bishop also posted some comparison data where the only battery related value there was vbat, with the values given in millivolt. From that and his manufacturerdata I deducted that the last wake byte has to be included, which then give the exact (milli)volt values as Bishop was stating.

Looking at everything again now, it seems that Bishop as well as @yoavf either have very low run down CR2032 batteries (values between 1.897 - 1.901 V), which could well be the case, or we still have an issue in the decoding.

If you have experience with this sensor it would be great if you can shed some light on the remaining voltage issue.

Thanks

jmtorres:
As I mentioned in the forum, I couldn’t quite correlate the battery voltage or percentage to what was being returned under vbat from the API. I tested with a new battery and I also have about 20 “new” Sony CR2032 batteries. I’ll try to get some actual readings with a VOM and compare to the BLE values and the API values when I have some time.

jmtorres:
Actually, I just remembered that the value being read from BLE did match what was in the API, so it should just be a matter of scaling/converting to a useful number, whether that's a percentage or voltage.

DigiH:
That would be great if you could get some readings with brand new batteries, then we at least know around which value the vbat should be and can look for a possibly more correct decoding for it :)

The API values would be great, but knowing the ~3.2V range of a new CR2032 just some data reading would already be helpful.

Thanks

Ernst79:
We can just try to reuse some volt to battery percentage conversion of some other sensors. There are more sensors that report in volt, where I convert it to battery.

DigiH:
What still confuses me are the cloud results which the user Bishop got, and I put in my comparison chart further up.

They match exactly, so I initially assumed they should be correct, but all the results do look very low.

Does anyone know if those cloud results from the official SmartDry gateway, or are they from another 3rd party decoder?

jmtorres:
Here are some of my findings today:

The "Wake" bytes are actually part of the battery level. The last 4 bytes are the battery level as an INT16 LE. This is reflected in the "vbat" value in the SmartDry Server API response. With a new battery at ~3.15v, this value is 2048. After that battery dropped to 3.05v, the value is 1995. With the previous battery at 2.963v, the battery value was 1908. I suspect that the battery level is being read as a 12-bit signed integer which would max out at 2048.

However, on occasion, the value for the battery level will be a much smaller value. This is also reflecting in the "vbat" value in the API response. This is why I had agreed that the scale for the battery level was much smaller as "DieKatzchen" from the forum had assumed. So far the only values that have been recorded on my device are 124, 125, and 130. They normally only appear when the shake value is low, but not always. I have not been able to correlate them with a cycle starting or ending. It may just be easiest to discard these values under a certain threshold if we cannot determine what they mean.

I've attached some responses from the API and a spreadsheet of the recorded BLE values from the past couple of weeks.

[SmartDryAPI-vbat2047.json.txt](https://github.com/custom-components/ble_monitor/files/9195434/SmartDryAPI-vbat2047.json.txt)
[SmartDryAPI-vbat124.json.txt](https://github.com/custom-components/ble_monitor/files/9195435/SmartDryAPI-vbat124.json.txt)
[SmartDry Values.xlsx](https://github.com/custom-components/ble_monitor/files/9195429/SmartDry.Values.xlsx)

DigiH:
Thanks a lot @jmtorres, the vbat level including the last 4 bytes was the first thing I realised, still not too sure with the lower values. I will go through your findings to see if there is more to find out.

The battery going down a lot is related to the **Wake** byte included in the battery calculation. When  wake is on, i.e. the sensor is active, the byte is **07**, when wak is off it's **00**, requiring a recalculation of the battery voltage. We have already provisioned for that. What still throws me a little is the **06** state @yoavf reported in his testing.

DigiH:
Ok, I think I got it

ff07 => 07ff = 2047 + 1055 = 3102 = **3.102** V Slightly off you approximated new battery ~3.15v
cb07 => 07cb = 1995 + 1055 = 3050 = **3.05** V spot on
7407 => 0774 = 1908 + 1055 = 2963 = **2.963** V spot on

So the magic conversion number seems to be adding 1055 ;) at lest from those given observations.

This would also put all the previously calculated very low values between  **1.897 - 1.901** V into am much more reasonable **2.9 - 3.0** V range.

DigiH:
So with the above wake byte very likely also containing other relevant bits - bit[7-1], which would require many adjusted calculations, probably best to just use the single battery byte with the following calculation

ff = 255 + 2847 = 3102 = 3.102 V Slightly off you approximated new battery ~3.15v
cb = 203 + 2847 = 3050 = 3.05 V spot on
74 = 116 + 2847 = 2963 = 2.963 V spot on

and only using bit[0] of the following byte to get the **wake** state.

While it would still be nice to find out what the other bits[7-1] might indicate, the above should give us correct readings for whatever they might be set to.

DigiH:
And also looking at your logs some more @jmtorres , I think we also need to re-evaluate our **shake** calculations, as there always is a **"shake": 0,**/**"shake": 1,** and **"shakeCount": 3,**

My initial finding here is, that shake is not the whole 2 bytes [8] and [9], but byte [9], and more likely again just bit[0] of byte[9] indicating the (current) **"shake"** state, whereas bytes [8] is the **"shakeCount"**

jmtorres:
At the moment, the format that I've mentioned is accurately reflected in the official API response. The 2 API responses that I uploaded were to show the swings of the vbat value. The vbat value reflects at least 12 bits of the data. Everything beyond the 12th bit has been empty in all of my readings. 

The shake values in the API also reflect what is being read from the 4 bytes as this value maxes out above 256 but below 512, it only seems to be using 9-10 bits. However, the remaining bits have also been empty in all of the data I have recorded. 

I can provide additional snapshots of the API responses if it would be helpful. Everything beyond temperature, humidity, shake, and battery appears to be derived/calculated from those values on the server or BT hub. Unless the device manufacturer was outputting the wrong information into their own API, I am hesitant to assume otherwise. The one glaring oddity I see is the 124/125/130 for battery which is also output in the vbat property in their API.

Another thing to note is that neither the App or official SmartThings integration show the battery status. The only thing SmartDry offered was an alert when the battery is low and needs to be replaced.

DigiH:
> Unless the device manufacturer was outputting the wrong information into their own API, I am hesitant to assume otherwise

Well, were those API outputs when you stated the **3.05v** and **2.963v** above? All I ever saw from other API reporting were far too low unrealistic voltage numbers. And if you say that SmartDry only ever offered a battery low warning, then the API vbat numbers could well be off. 

But those could only be achieved by adding the additional calculation I found above. If those readings you provided above are correctly corresponding to the voltages I'm pretty confident and happy with my last voltage calculations.

> I can provide additional snapshots of the API responses if it would be helpful.

I'm still certain that **bits[7-1]** of **byte[11]** indicate other settings, which for you seem to be all OFF, making them all possible `<statename>: 0, ` settings in your two text files.

With me also being confident about the shake status just being **bit[0]** of **byte [9]**, that also leaves **bits[7-1]** of **byte[9]** for other setting indications.

If at all possible, could you try setting as many settings to ON/ACTIVE in the SmartDry app which are currently `<statename>: 0, ` in your files, just to see if you get a different **byte[11]** to **07** and **00** and/or **byte [9]** with **00** and **01**?

That is if you are willing to test this out further ;) The only problem with this device is, that once their service has shut down it will be next to impossible to try and reverse engineer any more that we can get before that date.

jmtorres:
I'm not sure what the scaling for the battery value should be, but I believe that the last 4 bytes reflect what is being read for the battery value (except when it is returning 124, 125, 130, or other small values). I couldn't correlate another bit that indicates when these bytes reflect the battery level or the other values.

I recorded some API responses while shaking the sensor and while toggling some of the settings in the app with duplicate responses removed. I'm fairly confident that all setting bits are being handled on the server or hub and the sensor is not aware of these. 

[SmartDry API Responses.xlsx](https://github.com/custom-components/ble_monitor/files/9202456/SmartDry.API.Responses.xlsx)

I do see some shake values in the API above 512 so there is probably more than 9 bits for that value. I suspect that the  shake value is derived from a 3 axis accelerometer (
U3
), but the resolution of the values are quite low. I'm tempted to find my stereo microscope and try to read the markings on the chip to pull up a datasheet, but I feel I'm getting to far down the rabbit hole on this.

We will be running a dryer load in an hour or so. I'll record the API responses during that load and post them here. I'm polling the API once per second. Hopefully that is enough frequency without spamming the API too much.

DigiH:
> I'm not sure what the scaling for the battery value should be, but I believe that the last 4 bytes reflect what is being read for the battery value (except when it is returning 124, 125, 130, or other small values). I couldn't correlate another bit that indicates when these bytes reflect the battery level or the other values.

I still think it's only 2 bytes/1 8-bit byte for the battery ;) this also would not produce the small values you are experiencing, as those are the values when the following 1 8-bit byte (the **wake** byte) is 00.

My thinking that the other 7 of the 8 bits of that byte also representing other settings came with the observation from @yoavf, whose data always had **06** at the end in the testing, instead of your 00 when the sensor was lying still

```
ae01203fee4150d55f420100c706
ae0144e7ed41008f5f420100c706
ae017a63ed4140315f420100c706
ae01ca9dec4160fc5f420100c306
ae0124c2eb41000c60420200c406
ae017de6ea41702360420000c406
```

> SmartDry API Responses.xlsx

Looks interesting, especially the **Toggling settings** tab, but I cannot see any related BLE broadcast **manufacturerdata** , which would really make it interesting to compare with these new settings.

> I'm tempted to find my stereo microscope and try to read the markings on the chip to pull up a datasheet, but I feel I'm getting too far down the rabbit hole on this.

Yes, I don't think you should go that far ;) especially since the raw data could/might still be converted before being sent on.

> I'll record the API responses during that load and post them here. I'm polling the API once per second. Hopefully that is enough frequency without spamming the API too much. I'll also try to link the response to the associated BLE broadcasts, but that may be difficult.

The latter would be much more important than scanning every second, the same with the new speadsheet above. Only having the API data doesn't help much, if it's not related to the actual BLE broadcasts, like in your SmartDry Values.xlsx.

Thanks

jmtorres:
> I'm confident this happened when you had some of the settings changed to ON, as this was never seen before with any shake data, therefore it would confirm my theory. But without any related BLE broadcast data it is unfortunately of little use for investigation.

The higher shake values were read before changing any settings in the app. The toggling tab was all from a few minutes later. I was testing by alternating all axises to try and see if I could determine a pattern.

> The latter would be much more important than scanning every second, the same with the new speadsheet above. Only having the API data doesn't help much, if it's not related to the actual BLE broadcasts, like in your SmartDry Values.xlsx.

My son just had an emergency at summer camp so I wasn't able to capture the load that went in. I'll try to capture both when I do laundry this weekend. I'd like to gather as much information as necessary before the API is shut down. It is a bit difficult to correlate the values since there isn't an identifier or timestamp in the BLE data that I can reference. If you have any suggestions on how to capture that data, that would be helpful. I'm currently using my esphome ble_tracker to capture the data from the on_ble_manufacturer_data_advertise event. My ble_monitor is currently reading from the built-in BT on my HA's Raspberry Pi, but it's not close enough to the dryer to receive all the advertisements. I have another ESP that I could potentially set up as a ble gateway, but I haven't had time to mess around with it.

DigiH:
> My son just had an emergency at summer camp

I'm sorry to hear that, I hope everything is fine.

> I'd like to gather as much information as necessary before the API is shut down.

I think that should be our goal, so that possibly we can even go back later to verify or re-evaluate some findings.

I'd also like to ask you to turn ON as many settings as possible again, also  settings like `"alexaDeviceEnable"` and `"notifyAlexa"`, even if you don't have or use Alexa. This is best done when the SmartDry is just resting on your desk or anywhere else close to your Raspberry Pi, without any changing values for temperature, humidity and shake when you're capturing the manufacturerdata, like in your first SmartDry Values.xlsx. Just the raw data will be fine, no conversions needed for this.

This will either confirm my theory about other bits' states being broadcast, or it will disproof it. Either finding would be good, as it will bring us closer to fully understanding the broadcast data.

For catching additional activity data - whatever you did, and whichever set-up you used for creating the SmartDry Values.xlsx speadsheet was great. Maybe not do several runs on the same kind of drying, but one with a heavy load like lots of towels, one with normal mixed clothes, one with delicates ... so that the logs will show the different kind of drying behaviour and data. Then catching some of the API data at the same time, even with slight time stamp discrepancies should still give a good pictures or the relationships. Unfortunately I cannot think of a set-up where the ble monitor manufacturerdata would be totally in sync with the API data from the cloud.

Thanks again for all your effort.

DigiH:
> I have another ESP that I could potentially set up as a ble gateway, but I haven't had time to mess around with it.

If you have this separate ESP32, and with the release of our SmartDry decoder in Theengs Decoder and its subsequent integration into OpenMQTTGateway, you can install the OpenMQTTGateway `esp32dev-ble-datatest` binary, easiest from the web upload page, if you have the Google Chrome or Microsoft Edge browser.

https://docs.openmqttgateway.com/upload/web-install.html

This way you will get the raw **manufacturerdata** and the **decoded data** all in one MQTT message, like

```
"{“id”:“AA:BB:CC:DD:FF:4A”,“mac_type”:0,“name”:“xBBQ”,“manufacturerdata”:“ae01ca9dec4160fc5f424a005206”"brand":"SmartDry","model":"Laundry Sensor","model_id":"SDLS","tempc":29.57704544,"tempf":85.23868179,"hum":55.99645996,"shake":false,"shake_count":74,"volt":2.929,"wake":false}"
```

You can then use the MQTTX client and subscribe to the topic home/OpenMQTTGateway_ESP32_BLE/BTtoMQTT/AABBCCDDEEFF (AABBCCDDEEFF being the MAC address of your ESP32)

https://github.com/emqx/MQTTX

Possibly you can subscribe to a second topic to additionally log the incoming API data at the same time.

After a while, when having enough data collected, you can use the **Export Data** option in the top right … menu of MQTTX, to save it all to disk.

If the data is better displayed and then logged as **Plaintext** or **JSON** would be something to try out for the best and most concise result in the end.

I hope this helps in getting more data for current refinement of the decoders and for future re evaluations.

Thanks

PR #977, opened 2022-08-01
add xiaomi lock E

huexpub:


Ernst79:
Thanks, did you check wether the lock also sends the armed away ble data? You can check by adding "armed away" as binary sensor, like for the ZLMS17LM

huexpub:
emm, add "armed away" not effect ,butit should have this function, I will analyze later few days

Ernst79:
Did you analyse it?

Ernst79:
Are you sure about the model name. According to here, it is XMZNMS04LM. This number looks more in line with the others.

PR #1008, opened 2022-09-04
Update M1S-T500.md

lerra:
If the device is not setup with mi home app first it will only setup empty pay load and it is hard to troubleshoot this if there is no logs connected to this


Ernst79 (approved):

Ernst79:
Thanks,

PR #1012, opened 2022-09-07
Updated sv.json

el97:
Updated the Swedish translation and fixed some spelling. It would be nice if we could move over to poeditor for translating. It would be easier to translate since you got the strings in order without worrying that you miss something.


Ernst79:
Thanks.  will look into poeditor later

PR #1128, opened 2023-02-05
Add timeout without movement to xiaomi RTCGQ02LM

Ernst79:
Add object "timeout without movement" to Xiaomi parser (used by RTCGQ02LM)


CloCkWeRX (approved):
That seems.. .sensible.

Ernst79:
thx. This probably won't change much to the behavior, as BLE monitor already had its own timer to change the state to away. But it will at least get the confirmation messages of the away state from the sensor.

PR #1229, opened 2023-09-05
Update sk.json

misa1515:


Ernst79:
Thanks!. The test fails somehow over the word `Paket`, which is probably perfect in Slovakian, but not in English. 

I need to figure out how to fix that

Ernst79:
Ah, you already closed it

PR #1240, opened 2023-09-21
Make Thermopro advertisement parsing less strict to support TP357S

chq-matteo:
# tl;dr

The thermopro TP357S has 2 extra bytes in the ble message

# context

I bought a TP357S which didn't work out of the box with ble_monitor

But from the debug logs I could see that the current parsing works for humidity and temperature

e.g. in the debug logs:
```
2023-09-21 20:03:34.546 INFO (Thread-3) [custom_components.ble_monitor.ble_parser] BLE advertisement received from MAC/UUID xxxxxxxxxxxx: service data: []manufacturer specific data: [b'\x08\xff\xc2\xdd\x004"\x0b\x01']local name: TP357S (XXXX)UUID16: None,UUID128: None
```

The local name would match, but the manufacturer specific data is 9 bytes instead of 7

I couldn't figure out what the extra 2 bytes mean, but I know the bytes[3:6] are humidity and temperature as before

These are the results using the patch in this PR in home assistant

Screenshot 2023-09-21 202625

These are the values from the app (the values are slightly different because the second screenshot was taken a few minutes after)

photo_2023-09-21_20-28-54

### DIFF
--- a/custom_components/ble_monitor/ble_parser/__init__.py
+++ b/custom_components/ble_monitor/ble_parser/__init__.py
@@ -453,7 +453,7 @@ class BleParser:
                         # Inkbird IBS-TH
                         sensor_data = parse_inkbird(self, man_spec_data, local_name, mac, rssi)
                         break
-                    elif local_name[0:5] in ["TP357", "TP359"] and data_len == 0x07:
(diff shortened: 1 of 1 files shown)


Ernst79 (approved):
Thanks, lgtm

Issue #1285, opened 2024-01-06
Jaalee JHT sensor wrongly calculated temperature and humidity data in HA


gya352:
### What happened?

I recently purchased a new Jaalee JHT heat and humidity sensor. I found that the heat and humidity values shown by the app and HA differ significantly. The problem is that the calculation of the sensor data in the jaalee.py file seems to be incorrect. The correct calculation method can be found on the website below: https://sensor.jaalee.com/scan_api.html. A different calculation method is used in the  jaalee.py file.

### Sensor type

Jaalee JHT

Ernst79:
Thanks, the temperature formula was reverse engineered in the past, as we were not able to find any formula. I'll check if the formula's make sense and if they do, will implement them

Ernst79:
Thanks, released as 12.8.0

PR #1288, opened 2024-01-06
fix jalee calculations

Ernst79:
fix jalee calculations

See #1285


gya352 (approved):
The new calculation formula is good. The calculated values seems correct.

PR #1337, opened 2024-03-21
Add support for Acconeer XM126 distance sensor

jbjornstedt:


custom_components/ble_monitor/ble_parser/acconeer.py
```diff
@@ -8,11 +8,13 @@
 
 ACCONEER_SENSOR_IDS = {
     0x80: "Acconeer XM122",
+    0x90: "Acconeer XM126 Distance",
```
Ernst79, line 11: Is this a different device compared to the XM126?

jbjornstedt: The hardware is the same but the firmware on the device is different

Ernst79: Ok, I want to suggest to move the device to the AUTO_MANUFACTURER_DICT in const.py. Such that sensors are added when they are broadcasted. In that way we can use both firmware versions as XM126. I can make that change later this weekend.

Ernst79 (approved):

PR #1433, opened 2025-02-07
Downgrade error to warning

CloCkWeRX:
As this can be a transitory issue; downgrades the log from "error" (just won't ever work) to warning.


Ernst79 (approved):
thanks

PR #1470, opened 2025-06-06
Support: Beckett Genisys heater oil burner interface

jnj16180340:
Supports this pair of devices; others from the same manufacturer may be compatible:
- (BLE interface) https://www.beckettcorp.com/product/beckettlink-pro-7652a/
- (Host controller) https://www.beckettcorp.com/product/genisys-7505-120v-oil-burner-control/

Example of a typical burner cycle:
Screenshot_20250606_171419

Please let me know if you want any changes etc. - This has been running in my home for a week or so seemingly without issue


Ernst79 (changes requested):
Thanks!  
Looks good to me, however, there are some small pre commit errors that need to be fixed.

PR #1573, opened 2026-09-08
Handle unknown Grundfos pump modes

IoannisArmamentos:
## Summary

Handle unknown Grundfos MI401 pump mode values without crashing the BLE parser.

## Problem

The Grundfos parser used a direct lookup in `PUMP_MODE_DICT`.

If a device reports a pump mode that is not currently known by the integration, the lookup raises `KeyError`. Since the value comes directly from the BLE advertisement, this can also happen with malformed packets or future firmware introducing a new mode.

## Fix

Use a safe dictionary lookup and report unknown values as:

`Unknown (<value>)`

Known pump modes remain unchanged.

A regression test covers an advertisement containing pump mode `255`.


Ernst79 (approved):
thanks

PR #1574, opened 2026-09-08
Avoid mutating shared restore attribute lists

IoannisArmamentos:
## Summary

Avoid mutating the shared `RESTORE_ATTRIBUTES` lists when restoring entity state.

## Problem

Several restore paths assigned `RESTORE_ATTRIBUTES` directly to a local variable and then appended another attribute to it.

Because lists are mutable, this modified the shared module-level list in place.

The most significant occurrence was in `device_tracker.data_update()`, where the mutation could happen repeatedly during normal BLE updates and continuously add duplicate entries for the lifetime of Home Assistant.

The same pattern was also present in the one time restore paths for device tracker, sensor, and binary sensor entities.

## Fix

Create a new list containing the base restore attributes plus the entity-specific `mac_address` or `uuid` attribute instead of appending to the shared list.

Restore behavior and attribute ordering remain unchanged, while `RESTORE_ATTRIBUTES` itself is no longer modified.

### DIFF
--- a/custom_components/ble_monitor/binary_sensor.py
+++ b/custom_components/ble_monitor/binary_sensor.py
@@ -336,8 +336,7 @@ class BaseBinarySensor(RestoreEntity, BinarySensorEntity):
         elif old_state.state == STATE_OFF:
             self._state = False
 
-        restore_attr = RESTORE_ATTRIBUTES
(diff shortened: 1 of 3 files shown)


Ernst79 (approved):
Thanks

PR #1576, opened 2026-09-08
Guard invalid Xiaomi motion time payloads

IoannisArmamentos:
## Summary

Prevent malformed Xiaomi MiBeacon motion-time objects from raising an exception during parsing.

## Problem

`obj4850()` parses a one byte motiontime value using `struct.unpack("<B", xobj)`.

The payload length originates from the BLE advertisement and was not validated before unpacking. Payloads with an unexpected length could therefore raise `struct.error` and propagate out of the parser.

## Fix

Require `obj4850` payloads to contain exactly one byte before unpacking them.

Valid one byte payloads remain unchanged, while malformed payloads now return an empty result instead of raising an exception.

Regression tests cover both valid and invalid payload lengths.

### DIFF
--- a/custom_components/ble_monitor/ble_parser/xiaomi.py
+++ b/custom_components/ble_monitor/ble_parser/xiaomi.py
@@ -853,6 +853,9 @@ def obj484f(xobj):
 
 def obj4850(xobj):
     """Time in minutes with motion (not used, we use 484e)"""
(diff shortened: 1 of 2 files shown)


Ernst79 (approved):

Ernst79:
thanks

PR #1577, opened 2026-09-08
Handle invalid BTHome UTF-8 strings

IoannisArmamentos:
## Summary

Prevent malformed BTHome text measurements from raising `UnicodeDecodeError`.

## Problem

`parse_string()` decoded raw BLE payload bytes directly as UTF-8 without handling invalid byte sequences.

Because BTHome measurement data comes directly from BLE advertisements, malformed or unexpected string payloads could raise `UnicodeDecodeError` and propagate out of the parser.

## Fix

Catch invalid UTF-8 decoding errors and return `None`, which is the existing sentinel used by the BTHome measurement pipeline for values that cannot be parsed.

Valid UTF-8 strings remain unchanged.

Malformed string payloads are logged at debug level to avoid repeated invalid advertisements flooding Home Assistant logs.

A regression test covers invalid UTF-8 text payloads.

### DIFF
--- a/custom_components/ble_monitor/ble_parser/bthome.py
+++ b/custom_components/ble_monitor/ble_parser/bthome.py
@@ -43,9 +43,13 @@ def parse_float(data_obj: bytes, factor: float = 1.0):
     return round(val * factor, decimal_places)
 
 
-def parse_string(data_obj: bytes) -> str:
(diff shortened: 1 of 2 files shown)


Ernst79 (approved):

Ernst79:
Thanks, you are doing a great job!

PR #1578, opened 2026-09-08
Allow 100% humidity readings

IoannisArmamentos:
## Summary

Allow valid humidity readings of exactly 100% RH.

## Problem

The shared humidity spike filter currently uses an upper limit of `99.9`, causing valid `100.0%` humidity readings to be discarded.

This limit originates from the early single device version of the integration, where it matched the documented display range of the Xiaomi LYWSDCGQ/01ZM.

The BLE value itself is not limited to `99.9`, and the integration now supports many other humidity sensors with valid ranges extending to 100% RH.

## Fix

Increase the shared humidity upper bound from `99.9` to `100.0`.

Values above 100% remain rejected by the existing spike filter logic.


Ernst79 (approved):
thanks again

PR #1579, opened 2026-09-08
Guard empty Xiaomi button payloads

IoannisArmamentos:
## Summary

Prevent empty Xiaomi MiBeacon button event payloads from raising `IndexError`.

## Problem

The Xiaomi parsers for single click, double click, and long press events access the first payload byte for several device types.

These object types can legitimately reach the parser with an empty payload, so malformed advertisements for affected devices could trigger `xobj[0]` on an empty byte string and raise `IndexError`.

Some Xiaomi devices intentionally use an empty payload to signal a button event, so a blanket empty-payload guard would incorrectly break valid behavior.


Ernst79 (approved):
Thanks

PR #1580, opened 2026-09-08
Handle unknown PTX button values

IoannisArmamentos:
## Summary

Prevent unexpected PTX-F1-Display button values from causing a Xiaomi parser exception.

## Problem

The PTX-F1-Display branches in `obj4e0c()`, `obj4e0d()`, and `obj4e0e()` returned `None` when the click byte was not one of the recognized values.

The Xiaomi parser passes the returned value directly to `dict.update()`, so an unknown click value caused:

`TypeError: 'NoneType' object is not iterable`

The click byte originates from the BLE advertisement and can contain unexpected values.

## Fix

Return an empty result for unrecognized PTX-F1-Display click values instead of `None`.

Recognized button events remain unchanged.

Regression tests cover unknown and recognized PTX-F1-Display button values.

### DIFF
--- a/custom_components/ble_monitor/ble_parser/xiaomi.py
+++ b/custom_components/ble_monitor/ble_parser/xiaomi.py
@@ -1073,7 +1073,7 @@ def obj4e0c(xobj, device_type):
                 "button switch": "single press",
             }
         else:
-            result = None
+            result = {}
     else:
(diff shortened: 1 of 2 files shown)


Ernst79 (approved):
And again, thanks, much appreciated

PR #1582, opened 2026-09-08
Handle invalid Xiaomi button objects

IoannisArmamentos:
## Summary

Prevent invalid Xiaomi MiBeacon button objects from causing a parser exception.

## Problem

`obj1001()` returned `None` for invalid payload lengths and for unrecognized device types.

The Xiaomi parser passes object-parser results directly to `dict.update()`, so these cases could result in:

`TypeError: 'NoneType' object is not iterable`

Both the object length and device identifier originate from the BLE advertisement and can contain unexpected values.

## Fix

Return an empty result for invalid payload lengths and unsupported device types instead of `None`.

Valid button event parsing remains unchanged.

Regression tests cover invalid input and a representative valid button event.


Ernst79 (approved):
thanks

PR #1592, opened 2026-09-08
Respect Bluetooth interface validation errors

IoannisArmamentos:
## Summary

Ensure invalid Bluetooth interface selections are rejected by the config and options flows.

## Problem

Both the initial configuration flow and the options flow could set the
`cannot_disable_bt_interface` validation error when `disable` was selected together
with another Bluetooth adapter.

However, both flows continued to create or save the entry instead of redisplaying the
form with the validation error.

## Fix

Return the main form whenever validation errors are present before creating or saving
the entry.

Valid Bluetooth interface selections continue to follow the existing behavior unchanged.


Ernst79 (approved):
LGTM

PR #1593, opened 2026-09-09
Expose consumable sensor for M1S-T500

IoannisArmamentos:
## Summary

Expose the consumable percentage reported by Xiaomi M1S-T500 toothbrushes.

## Problem

The Xiaomi parser already handles object `0x1013` and produces a `consumable` value, and BLE Monitor already provides a generic consumable sensor type.

However, the M1S-T500 model mapping did not include the `consumable` measurement, so the parsed value was never exposed as a Home Assistant entity.


Ernst79 (approved):
LGTM

PR #1597, opened 2026-09-09
Validate device type in Mocreo ST7 advertisements

IoannisArmamentos:
## Summary

Reject malformed Mocreo ST7 advertisements whose embedded device type does not match the ST7 packet format.

## Problem

Mocreo uses different advertisement layouts for ST7 and other devices.

The special ST7 format uses a different payload offset, but a corrupted device-type byte could cause that shorter payload to be interpreted as an SD1 advertisement.

The SD1 parsing path then accessed fields that are not present in the ST7 layout, resulting in `IndexError`.


Ernst79 (approved):
Thanks

PR #1600, opened 2026-09-10
Expose no motion duration for MJYD02YL

IoannisArmamentos:
## Summary

Expose the idle/no-motion duration reported by Xiaomi MJYD02YL devices as a standalone sensor.

## Problem

The Xiaomi parser already decodes object `0x1017` and reports the elapsed no-motion time in seconds.

However, MJYD02YL did not map this parsed value to a Home Assistant sensor, so the information was discarded.

## Fix

Add a standalone duration sensor for the existing no motion time value and map it to MJYD02YL.

The sensor reports the number of seconds since motion was last detected.

Existing motion, light, battery, RSSI, reset timer, and parser behavior remains unchanged.

Regression coverage verifies the known `0` and `120` second values while preserving the existing motion state behavior.


Ernst79 (approved):
Lgtm, thanks

PR #1601, opened 2026-09-10
Expose stabilized state for Mi Scale

IoannisArmamentos:
## Summary

Expose the Mi Scale stabilization state as a standalone binary sensor.

## Problem

The Mi Scale parser already reports whether a measurement has stabilized, but BLE Monitor only exposed this value as an attribute of the weight sensor.

This makes the stabilization state less convenient to use directly in dashboards and automations.

## Fix

Add a standalone binary sensor for the existing `stabilized` value and map it to both Mi Scale V1 and Mi Scale V2 devices.


Ernst79 (approved):

Ernst79:
Thanks

PR #1602, opened 2026-09-10
Ignore unknown Xiaomi door events

IoannisArmamentos:
## Summary

Prevent unknown Xiaomi door event values from generating false closed states.

## Problem

The Xiaomi `0x1019` parser supports the known door-event values `0` through `3`.

However, any unsupported value previously fell through to a default result with
`opening = 0`, which could incorrectly report the door as closed.

## Fix

Ignore unsupported `0x1019` event values instead of generating a door state update.


Ernst79 (approved):
LGTM

PR #1611, opened 2026-09-20
Fix iNode Care Sensor temperature decoding

### LINKED ISSUE #1610: [Bug]: TypeError in inode.py for iNode models 0x92 and 0x9A
### What happened?

There is a bug in `ble_parser/inode.py` when parsing temperature data for iNode device types `0x92` and `0x9A`.

The `raw_t` value returned by:

```python
unpack("<HHHHHHHQ", xvalue)
```

is an `int`. However, the parser attempts to index it:

```python
msb = raw_t[0]
lsb = raw_t[1]
```

This raises:

```text
TypeError: 'int' object is not subscriptable
```

and prevents the advertisement from being parsed.

The issue is currently not detected by `test_inode_parser.py`, because these two device types do not appear to have dedicated test cases.

### Proposed solution

Extract the two bytes using bitwise operations. The variable names below also reflect the actual byte order of the little-endian integer:

```python
low_byte = raw_t & 0xFF
high_byte = (raw_t >> 8) & 0xFF

temp = low_byte * 0.0625 + 16 * (high_byte & 0x0F)
if high_byte & 0x10:
    temp -= 256
```

A regression test for device types `0x92` and `0x9A` should also be added.

### Sensor type

iNode Care Sensor 2 (0x92), iNode Care Sensor T (0x9A)

### Relevant log output

```shell

```

dajiaohuang:
Fixes #1610.\n\n## Summary\n- Decode the raw iNode Care Sensor 2 and Care Sensor T temperature word using its little-endian byte layout.\n- Add regression coverage for both device IDs.\n\n## Validation\n- pre-commit run --files custom_components/ble_monitor/ble_parser/inode.py custom_components/ble_monitor/test/test_inode_parser.py (pass)\n- python -m py_compile custom_components/ble_monitor/ble_parser/inode.py custom_components/ble_monitor/test/test_inode_parser.py (pass)\n- pytest was not runnable locally because the pinned pytest-homeassistant-custom-component 0.13.354 requires Python 3.14 and this host has Python 3.13 active.


Ernst79 (approved):

Ernst79:
Thanks

Ernst79 (changes requested):
The test now fail, can you please check if is is intentionally or not?

dajiaohuang:
The failing fixture was using raw temperature

Ernst79 (approved):

